PR #1, opened 2015-02-24
Add create record of guidance feature @wipped

gjvis:
A very high-level "walking skeleton" feature, to allow us to keep moving.

What do people think?


stevenwilkin:
:+1:

PR #2, opened 2015-02-25
Capture guidance

stevenwilkin:
First cut of capturing the details of the delivered guidance. Just capturing the attendee's name currently.

### DIFF
--- a/features/step_definitions/record_of_guidance_steps.rb
+++ b/features/step_definitions/record_of_guidance_steps.rb
@@ -1,0 +1,8 @@
+When(/^appointment details are captured$/) do
+  page = AppointmentSummaryPage.new
+  page.load
+  page.name.set 'Joe Bloggs'
+  page.submit.click
+
+  expect(AppointmentSummary.count).to eql(1)
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,7 @@ AllCops:
   RunRailsCops: true
   Exclude:
     - 'bin/*'
+    - 'db/*'
     - 'vendor/**/*'
 
 Documentation:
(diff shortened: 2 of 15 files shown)


andrewgarner:
How about `GuidanceRecord` instead of `Guidance`?

andrewgarner:
Or maybe `AppointmentRecord`?

stevenwilkin:
Any opinions @gjvis ?

features/support/pages/guidance_page.rb
```diff
@@ -0,0 +1,6 @@
+class GuidancePage < SitePrism::Page
+  set_url '/guidance/new'
+
+  element :name, 'input.t-name'
+  element :submit, 'input[type="submit"]'
```
andrewgarner: We mixing our approach here. We're using test CSS classes and CSS attribute selectors. Why not `input.t-submit`?

features/step_definitions/record_of_guidance_steps.rb
```diff
@@ -0,0 +1,8 @@
+When(/^appointment details are captured$/) do
+  @page = GuidancePage.new
```
andrewgarner: We don't need an instance of the page, we're only using it within this method.

stevenwilkin: Good point. My first thought was that we'd need access to it in the other step definitions but on looking at them again, the rest of them don't involve the guidance capture form

features/support/pages/guidance_page.rb
stevenwilkin: Sure, I've no argument not to use CSS classes to target all the elements on the page

stevenwilkin:
Model renamed to ``AppointmentSummary`` per discussion

stevenwilkin:
RuboCop!

PR #5, opened 2015-02-26
Capture email address

andrewgarner:


app/views/appointment_summaries/new.html.erb
```diff
@@ -7,7 +7,13 @@
...
 <%= form_for @appointment_summary, url: appointment_summaries_path do |f| %>
-  <%= f.label :name %>
-  <%= f.text_field :name, class: 't-name' %>
+  <div class="l-form-row">
+    <%= f.label :name %>
+    <%= f.text_field :name, class: 't-name' %>
+  </div>
+  <div class="l-form-row">
+    <%= f.label :email_address %>
+    <%= f.text_field :email_address, class: 't-email_address' %>
```
stevenwilkin: `t-email_address` or `t-email-address` considering we have `l-form-row` above it

features/step_definitions/record_of_guidance_steps.rb
```diff
@@ -2,7 +2,11 @@
   page = AppointmentSummaryPage.new
   page.load
   page.name.set 'Joe Bloggs'
+  page.email_address.set 'joe.bloggs@example.com'
   page.submit.click
 
-  expect(AppointmentSummary.count).to eql(1)
+  appointment_summary = AppointmentSummary.last
```
gjvis: There is a race condition here (test runner vs application), but its probably fine for now.

PR #6, opened 2015-02-26
Pdf

stevenwilkin:


features/step_definitions/record_of_guidance_steps.rb
```diff
@@ -6,3 +6,9 @@
 
   expect(AppointmentSummary.count).to eql(1)
 end
+
+Then(/^a record of guidance document is created$/) do
+  expect(page.response_headers['Content-Type']).to eql('application/pdf')
+  text = PDF::Inspector::Text.analyze(page.source).strings.join
```
gjvis: Race condition here (test-runner vs application), we can sort it out in a future PR if it becomes an issue.

features/step_definitions/record_of_guidance_steps.rb
```diff
@@ -6,3 +6,9 @@
...
+  expect(text).to include('Joe Bloggs')
```
gjvis: Race condition here (test-runner vs application), we can sort it out in a future PR if it becomes an issue.

stevenwilkin:
Rebased

gjvis:
:+1: :ship:

PR #9, opened 2015-02-27
Email output document

stevenwilkin:
The email body is currently empty but there is a card on the wall for this.

I'm not 100% convinced on the usage of Cucumber here to verify emailing of the document but it's a starting point for discussion.

Another thing to talk about is potentially using factories as the 2 spec files introduced here manually create instances of `AppointmentSummary` which will need maintained as the model grows.

### DIFF
--- a/app/domain/output_document.rb
+++ b/app/domain/output_document.rb
@@ -1,0 +1,16 @@
+class OutputDocument
+  include Rails.application.routes.url_helpers
+
+  attr_accessor :appointment_summary
+
+  def initialize(appointment_summary)
+    @appointment_summary = appointment_summary
+  end
+
+  def content
+    app = ActionDispatch::Integration::Session.new(Rails.application)
+    path = appointment_summary_path(appointment_summary, format: :pdf)
+    app.get(path)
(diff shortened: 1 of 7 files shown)


app/domain/output_document.rb
```diff
@@ -0,0 +1,16 @@
+class OutputDocument
```
andrewgarner: `AppointmentSummaryDocument`?

app/mailers/output_document_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class OutputDocumentMailer < ActionMailer::Base
```
andrewgarner: `AppointmentSummaryMailer`?

stevenwilkin: Yeah, talking about "appointment summaries" rather than "output documents" probably makes more sense

stevenwilkin:
I've updated `spec/mailers/output_document_mailer_spec.rb` to no longer use the database. `spec/domain/output_document_spec.rb` still needs it though.

PR #11, opened 2015-02-27
Email output document - redux

stevenwilkin:
Another go at #9.

The Cucumber stuff is a bit ropey.

### DIFF
--- a/app/controllers/appointment_summaries_controller.rb
+++ b/app/controllers/appointment_summaries_controller.rb
@@ -6,6 +6,7 @@ class AppointmentSummariesController < ApplicationController
   def create
     @appointment_summary = AppointmentSummary.create(appointment_summary_params)
     if @appointment_summary.persisted?
+      IssueOutputDocument.new(@appointment_summary).call
       redirect_to appointment_summary_path(@appointment_summary, format: :pdf)
     else
       render :new
--- a/app/mailers/output_document_mailer.rb
+++ b/app/mailers/output_document_mailer.rb
@@ -1,0 +1,10 @@
+class OutputDocumentMailer < ActionMailer::Base
+  def issue(appointment_summary, output_document)
+    attachments['record_of_guidance.pdf'] = output_document
+    mail(
+      to: appointment_summary.email_address,
+      from: 'guidance@pensionwise.gov.uk',
+      subject: 'Record of your guidance'
+    )
+  end
+end
--- a/app/services/issue_output_document.rb
+++ b/app/services/issue_output_document.rb
@@ -1,0 +1,15 @@
+class IssueOutputDocument
+  attr_accessor :appointment_summary
+
+  def initialize(appointment_summary)
+    @appointment_summary = appointment_summary
+  end
+
+  def call
+    return unless @appointment_summary.email_address.present?
+
+    output_document = OutputDocument.new(@appointment_summary)
+    OutputDocumentMailer.issue(@appointment_summary, output_document.pdf)
+      .deliver_later
+  end
+end
--- a/features/step_definitions/record_of_guidance_steps.rb
+++ b/features/step_definitions/record_of_guidance_steps.rb
@@ -1,4 +1,6 @@
 When(/^appointment details are captured$/) do
+  allow(OutputDocument).to receive(:new).and_return(double(pdf: ''))
+
   page = AppointmentSummaryPage.new
   page.load
   page.name.set 'Joe Bloggs'
@@ -13,6 +15,8 @@ end
 
 Then(/^a record of guidance document is created$/) do
   expect(page.response_headers['Content-Type']).to eql('application/pdf')
-  text = PDF::Inspector::Text.analyze(page.source).strings.join
-  expect(text).to include('Joe Bloggs')
+end
+
+Then(/^emailed to the customer$/) do
+  expect(ActiveJob::Base.queue_adapter.enqueued_jobs.first[:args][0]).to eql('OutputDocumentMailer')
 end
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,6 +1,7 @@
 require 'capybara'
 require 'capybara/cucumber'
 require 'cucumber/rails'
+require 'cucumber/rspec/doubles'
 require 'site_prism'
 
 ActionController::Base.allow_rescue = false
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -39,4 +39,6 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  config.active_job.queue_adapter = :test
 end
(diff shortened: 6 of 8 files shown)


features/support/env.rb
```diff
@@ -1,9 +1,12 @@
...
 require 'cucumber/rails'
+require 'cucumber/rspec/doubles'
 require 'site_prism'
 
 ActionController::Base.allow_rescue = false
 DatabaseCleaner.strategy = :transaction
 
 Cucumber::Rails::Database.javascript_strategy = :truncation
+
+Rails.application.config.active_job.queue_adapter = :test
```
andrewgarner: I'd expect to see this in the `config/environments/test.rb`

stevenwilkin: Good spot!

app/controllers/appointment_summaries_controller.rb
```diff
@@ -15,6 +15,7 @@ def create
   def show
     appointment_summary = AppointmentSummary.find(params[:id])
     output_document = OutputDocument.new(appointment_summary)
+    OutputDocumentMailer.guidance_record(appointment_summary).deliver_later
```
andrewgarner: Firstly - why are we passing in `appointment_summary` not `output_document`?

Secondly - I still think this is a bit confused. 'Output document' and 'Guidance record' are the same thing - we should settle on one.

The specific email is saying 'here is your record for the first time'. I can envisage this may grown in the future to include further emails for things like 'here is your re-issued record of guidance'.

Maybe:

```rb
OutputDocumentMailer.issue(output_document).deliver_later

# or 

OutputDocumentMailer.first_issue(output_document).deliver_later

# or 

OutputDocumentMailer.initial_issue(output_document).deliver_later
```

stevenwilkin: The issue of passing in `appointment_summary` versus `output_document` was something @gjvis and myself discussed briefly on Friday.

The mailer currently requires the `AppointmentSummary` as it contains the email address to send the mail to and when we get the copy for the email body it'll likely require personal details from this object as well.

But yeah, instantiating the document twice, in both the controller and the mailer is a thing to think about. Passing the generated document into the mailer along with the summary is probably the way forward.

stevenwilkin: I wasn't happy with the method name here either, `guidance_record` was the only thing that seemed to fit at the time. I like `issue` much more.

gjvis: I'd be much happier with the mailer not having dependencies on other models. Named "value" parameters ftw. Potentially this means that we need the concept of a PDF?

stevenwilkin:
Updated:

* renamed mailer method
* generated output document is passed into mailer

app/controllers/appointment_summaries_controller.rb
```diff
@@ -15,11 +15,13 @@ def create
   def show
     appointment_summary = AppointmentSummary.find(params[:id])
     output_document = OutputDocument.new(appointment_summary)
+    output_document_pdf = output_document.pdf
+    OutputDocumentMailer.issue(appointment_summary, output_document_pdf).deliver_later
```
andrewgarner: The email address is optional. We need to put a guard in place to make sure the mailer is only triggered if we have an email address to send to.

gjvis: Why are we sending an email from the show action?

stevenwilkin: Good point about the action @gjvis , this is redirected to after the `AppointmentSummary` is created so perhaps it's better in the create action or elsewhere

features/step_definitions/record_of_guidance_steps.rb
```diff
@@ -13,6 +15,8 @@
 
 Then(/^a record of guidance document is created$/) do
   expect(page.response_headers['Content-Type']).to eql('application/pdf')
-  text = PDF::Inspector::Text.analyze(page.source).strings.join
-  expect(text).to include('Joe Bloggs')
+end
+
+Then(/^emailed to the customer$/) do
+  expect(ActiveJob::Base.queue_adapter.enqueued_jobs.first[:args][0]).to eql('OutputDocumentMailer')
```
stevenwilkin: Horrible!

stevenwilkin:
Updated once more with a service object to handle issuing of output document after the appoint summary details are captured

app/mailers/output_document_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class OutputDocumentMailer < ActionMailer::Base
+  def issue(appointment_summary, output_document)
```
andrewgarner: The `appointment_summary` argument implies a dependency on an object of type `AppointmentSummary`.

You could name it something that implies duck typing, e.g. `recipient_details` though given we only need an email address, just have that as the param?

app/services/issue_output_document.rb
```diff
@@ -0,0 +1,23 @@
...
+    OutputDocumentMailer.issue(@appointment_summary, output_document.pdf)
+      .deliver_later
+  end
+
+  private
+
+  def valid_email_address?
+    @appointment_summary.validate
+    @appointment_summary.email_address.present? &&
+      @appointment_summary.errors[:email_address].blank?
```
andrewgarner: Why do we need to perform validation here? It's the concern of `AppointmentSummary` to ensure records created are valid? Shouldn't we trust it?

app/mailers/output_document_mailer.rb
stevenwilkin: It's a tough call at the moment seeing as the body of the sent email is currently blank. Additional attributes may or may not be required from the `AppointmentSummary` further down the line

andrewgarner: Fair point. Do we have an email template that can expose those needs?

app/services/issue_output_document.rb
stevenwilkin: Good point. It's perhaps overkill checking the validation again here. Just checking if the email address is non-empty could be enough.

app/mailers/output_document_mailer.rb
stevenwilkin: Not that I'm aware of. @aduggin ?

features/step_definitions/record_of_guidance_steps.rb
andrewgarner: You could introduce https://github.com/gocardless/rspec-activejob as a dependency?

stevenwilkin: I had a look at it but all it really gives us is an `enqueue_a` matcher, not something that can be used to make an assertion about the jobs queued up.

We could go full yak shave and *write* a matcher :imp:

andrewgarner: Steady on!

stevenwilkin:
Have removed validation from the service object

andrewgarner:
:+1:

stevenwilkin:
Yay!

stevenwilkin:
I'm still not happy about the `ActiveJob` assertion in this PR but I think we can merge this now?

gjvis:
I think its good enough to keep us moving. We can restructure when we know more...

PR #18, opened 2015-03-04
Configure MVP authentication

andrewgarner:
- Requires a user account to generate a record of guidance
- User account creation is through the console by a sysadmin
- New accounts must be confirmed
- Failed login attempts will cause the account to be locked
- Authenticated sessions will expire after a set period of time


stevenwilkin:
Looks good, I'll write a quick test to verify that unauthenticated users are redirected to the login

PR #22, opened 2015-03-05
Collect postal address of appointment attendee

andrewgarner:
I've reviewed the GDS Service Manual guidance on addresses here - https://www.gov.uk/service-manual/user-centred-design/resources/patterns/addresses

Opted for the simplest approach initially which is a text box to collect the address. We may need to break this down into individual address lines later on when we establish requirements with the print provider.

**Needs to be rebased once #21 has been merged.**

### DIFF
--- a/app/assets/stylesheets/base/_forms.css.scss
+++ b/app/assets/stylesheets/base/_forms.css.scss
@@ -20,6 +20,10 @@ input[type="date"] {
   padding: .5em;
 }
 
+textarea {
+  padding: .5em;
+}
+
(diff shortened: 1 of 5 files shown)


stevenwilkin:
Looks good. Only thing to keep in mind is that database migrations happen *after* application deployment but in this instance we can likely get away with manually restarting the app for the new schema to take effect.

stevenwilkin:
I should have looked at #21 beforehand, ignore what I said about the db ;)

PR #32, opened 2015-03-11
Collect reference number when recording appointment summary

stevenwilkin:
image


app/models/appointment_summary.rb
```diff
@@ -10,4 +10,5 @@ class AppointmentSummary < ActiveRecord::Base
   validates :income_in_retirement, inclusion: { in: %w(pension other) }
   validates :guider_name, presence: true
   validates :guider_organisation, inclusion: { in: %w(tpas cab) }
+  validates :reference_number, presence: true
```
andrewgarner: The database field is an integer so it would make sense to validate the numericality too.

stevenwilkin: Good call

stevenwilkin:
Updated with validation to ensure the entered string parses as a number. This is intended to be merged in `master` now so 9ac53fa may need to be cherry-picked.

stevenwilkin:
@aduggin will drop in the changes to the output document to include the reference number

PR #39, opened 2015-03-18
Remove PDF generation

stevenwilkin:


app/controllers/appointment_summaries_controller.rb
```diff
@@ -20,14 +18,7 @@ def show
...
 
-    respond_to do |format|
-      format.html { render html: output_document.html.html_safe }
-      format.pdf do
-        send_data output_document.pdf,
-                  filename: 'pension_wise.pdf', type: 'application/pdf',
-                  disposition: :inline
-      end
-    end
+    render html: output_document.html.html_safe
```
andrewgarner: Do we want to render a preview of the actual document or just a confirmation view?

If we do we should defiantly extract the template into a shareable bower module.

stevenwilkin: My next PR will include a preview/confirmation page

PR #41, opened 2015-03-18
Collect address fields separately

andrewgarner:


app/views/appointment_summaries/new.html.erb
```diff
@@ -32,13 +32,41 @@
...
-  <div class="row">
-    <div class="form-group col-xs-6">
-      <%= f.label :address, 'Postal address' %>
-      <%= f.text_area :address, cols: 40, rows: 6, class: %w(form-control t-address) %>
-    </div>
+  <h3>Postal address</h3>
+
+  <div class="form-group">
+    <%= f.label :address_line_1,
+                'Building and street <span class="visuallyhidden">line 1</span>'.html_safe %>
```
gjvis: perhaps the block-style call is neater given the html?
```erb
 <%= f.label :address_line_1 do %>
  Building and street <span class="visuallyhidden">line 1</span>
<% end %>
```
Not particularly fussed either way.

gjvis:
:+1:

PR #42, opened 2015-03-18
Add migration to add name fields

andrewgarner:


stevenwilkin:
Is `name` to be dropped after this?

andrewgarner:
That's the plan yes. Same goes for `address`.

gjvis:
:+1:

PR #45, opened 2015-03-18
Collect distinct name fields

andrewgarner:


spec/domain/output_document_spec.rb
```diff
@@ -16,6 +20,6 @@
   describe '#html' do
     subject { output_document.html }
 
-    it { is_expected.to include(name) }
+    it { is_expected.to include('Mr Joe Bloggs') }
```
stevenwilkin: Perhaps compare against `"#{title} #{first_name} #{last_name}"` ?

PR #46, opened 2015-03-19
Remove email address from appointment summaries

andrewgarner:
This is no longer needed as all output will be issued as white mail.


gjvis:
:ship:

PR #48, opened 2015-03-19
Require strong passwords

andrewgarner:
Set fairly conservatively but will reject simple passwords.


stevenwilkin:
If a password is deemed not string enough is it obvious how to make it stronger?

andrewgarner:
The error message reads "Consider adding a number, symbols or more letters to make it stronger."

gjvis:
We should also add some descriptive text to the "change password" views.

andrewgarner:
I've added the missing devise views for invitations (the only point passwords are set).

I've added some tips on how to create a strong password.

gjvis:
on :fire: 

:+1:

PR #55, opened 2015-03-19
Add predicates to determine the type of guidance provided

stevenwilkin:
These will be used to determine which variation of the output document to create


gjvis:
I think this is fine, its hard to tell without the code that uses these methods existing yet.

:+1:

PR #56, opened 2015-03-19
Add different pages to output document depending on type of guidance

stevenwilkin:
Intended for use with the preview/confirmation page


stevenwilkin:
So this was the placeholders for the difference output document pages that was demoed @aduggin if you want to give the go ahead to this PR

gjvis:
This is now redundant

PR #58, opened 2015-03-20
Only apply certain validation for eligible appointment attendees

andrewgarner:
If the attendee does not have a DC pension we don't need to know about their pot size or source of income.

cc/ @aduggin


gjvis:
Looks good :+1:

PR #65, opened 2015-03-23
Add Google Tag Manager code

andrewgarner:
Depends on a fork of the GOV.UK admin template. I'd like to get this merged in then update our dependencies back to a Ruby gem when one is published.


stevenwilkin:
Seems legit

PR #72, opened 2015-03-26
Add placeholder for CSV representation of output docs

stevenwilkin:
This interface ties together the card dealing with batch processing and the one concerning the CSV generation itself


gjvis:
:+1:

PR #73, opened 2015-03-26
Basic structure of CSV generation from output document batch processing

stevenwilkin:
Well gents, what do you think?

### DIFF
--- a/app/domain/batch_process.rb
+++ b/app/domain/batch_process.rb
@@ -1,0 +1,17 @@
+class BatchProcess
+  attr_reader :output_documents
+
+  def initialize(output_documents)
+    @output_documents = output_documents
+  end
+
+  def csv
+    [headers].concat(output_documents.map(&:csv)).join
+  end
+
+  private
+
+  def headers
+    OutputDocument::CSVRenderer::FIELDS.join('|') + "\n"
+  end
+end
--- a/spec/domain/batch_process_spec.rb
+++ b/spec/domain/batch_process_spec.rb
@@ -1,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe BatchProcess do
+  let(:headers) { OutputDocument::CSVRenderer::FIELDS.join('|') }
+  let(:csv) { 'csv' }
+  let(:output_document) { double(csv: csv) }
+
+  describe '#csv' do
+    subject(:lines) { described_class.new([output_document]).csv.split("\n") }
+
+    specify { expect(lines.first).to eq(headers) }
+    specify { expect(lines.last).to eq(csv) }
+  end
+end


stevenwilkin:
Updated this to reflect latest changes on `master`

app/domain/batch_process.rb
```diff
@@ -0,0 +1,17 @@
+class BatchProcess
+  attr_reader :output_documents
+
+  def initialize(output_documents)
+    @output_documents = output_documents
+  end
+
+  def csv
+    [headers].concat(output_documents.map(&:csv)).join
```
andrewgarner: I'm not quite sure about this. We are trusting (we know) that `output_document.csv` is implemented using `OutputDocument::CSVRenderer`.

If `output_document` returned a hash we could keep all knowledge of how to generate a CSV private to this class and also use the standard library to generate the CSV.

app/domain/batch_process.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  private
+
+  def headers
+    OutputDocument::CSVRenderer::FIELDS.join('|') + "\n"
```
andrewgarner: This has potential to catch is out. We're trusting that fields won't need escaping can just be concatenated. Why not use http://ruby-doc.org/stdlib-2.2.1/libdoc/csv/rdoc/index.html ?

stevenwilkin: I had thought the headers would need some sort of special treatment but now that you mention it I think treating them just like another row would be the thing to do per [CSVRenderer#render](https://github.com/guidance-guarantee-programme/output/blob/master/app/domain/output_document/csv_renderer.rb#L20-L22)

gjvis: Perhaps move this to `CSVRenderer.header_row`

app/domain/batch_process.rb
gjvis: how about something like:
```ruby
def csv
  [headers].concat(output_document.map(&method(:csv_row))
end

def csv_row(output_document)
  OutputDocument::CSVRenderer.new(output_document).render
end
```

gjvis: or move this entire class to be a class method on CSVRenderer:

```ruby
def csv(output_documents)
  csv = []
  csv << header_row
  output_documents.each do |doc|
    csv << new(doc).render
  end

  csv
end
``

gjvis: or rename this file `CSVRenderer`, and rename the current `CSVRenderer` to `CSVRowRenderer`

andrewgarner: With `header_row ` we still know that `output_documents` are rendered by `CSVRenderer`.

stevenwilkin:
Per discussion, I'm going to close this PR and rework this class into a `CSVRenderer` which then makes use of `OutputDocument::CSVRowRenderer`

app/domain/batch_process.rb
gjvis: well yes, something has to, this is that thing.

stevenwilkin:
The batch process will then be orchestrated by another object

PR #75, opened 2015-03-27
Record whether pension pots value is approximate

gjvis:


andrewgarner:
The code depends on the migration, but the code is deployed first then the database is migrated.

We need to break the migration out into a separate PR and release that first.

stevenwilkin:
:+1:

gjvis:
doh, sorry, rushing too much this morning. will split into two!

PR #82, opened 2015-03-30
Take into consideration approximate pension pot values

stevenwilkin:


app/domain/output_document.rb
```diff
@@ -101,4 +101,10 @@ def template_for(section)
     File.read(
       Rails.root.join('app', 'templates', "output_document_#{section}.html.erb"))
   end
+
+  def to_currency(number_string)
+    return '' if number_string.blank?
+
+    number_to_currency(number_string, precision: 0)
```
andrewgarner: `number_to_currency` takes `number` as an argument, `number_string` is potentially misleading.

app/domain/output_document.rb
```diff
@@ -33,15 +33,15 @@ def guider_organisation
...
-
-    pension_pot = number_to_currency(appointment_summary.value_of_pension_pots, precision: 0)
-
-    if appointment_summary.upper_value_of_pension_pots.blank?
-      pension_pot
-    else
-      pension_pot + ' to ' + \
-        number_to_currency(appointment_summary.upper_value_of_pension_pots, precision: 0)
+    pension_pot = to_currency(appointment_summary.value_of_pension_pots)
+    upper_limit = to_currency(appointment_summary.upper_value_of_pension_pots)
```
andrewgarner: These variable names could be a bit confusing in the future to someone not quite so absorbed as we currently are in this domain.

`approximation` not `approximate`?

We take an upper value then call it a limit? Maybe `upper_value`? Then maybe `lower_value`?

gjvis: I've updated the approximate variable. Tried changing the others but it ends up a bit harder to parse the method (you've got to read the whole thing quite carefully when pension_pot is referred to as `lower_limit` or `lower_value`).

Issue #79, opened 2015-03-30
Implement OutputDocument#greeting


gjvis:
**OutputDocument#greeting**:
```ruby
def greeting
  "You recently had a Pension Wise guidance appointment with #{guider_name} from #{guider_organisation} on #{appointment_date}"
end
```
(+ tests)

stevenwilkin:
We are now referring to this as `lead` and it's been implemented in #84

Issue #81, opened 2015-03-30
Format dates consistent with the GOV.UK service manual


gjvis:
`OutputDocument#appointment_date` should format dates according to the GOV.UK service manual, see
https://www.gov.uk/service-manual/user-centred-design/resources/patterns/dates.html#formatting-dates

stevenwilkin:
Resolved with #86

PR #83, opened 2015-03-30
Extract `greeting` into `OutputDocument`

stevenwilkin:
Addresses #79

### DIFF
--- a/app/domain/output_document.rb
+++ b/app/domain/output_document.rb
@@ -37,11 +37,16 @@ class OutputDocument
 
     pension_pot = number_to_currency(appointment_summary.value_of_pension_pots, precision: 0)
 
-    if appointment_summary.upper_value_of_pension_pots.blank?
-      pension_pot
+    unless appointment_summary.upper_value_of_pension_pots.blank?
+      upper_limit = number_to_currency(
+        appointment_summary.upper_value_of_pension_pots, precision: 0)
+      return "#{pension_pot} to #{upper_limit}"
+    end
+
+    if appointment_summary.value_of_pension_pots_is_approximate?
+      "#{pension_pot} (approximately)"
     else
-      pension_pot + ' to ' + \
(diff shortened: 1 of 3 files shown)


app/domain/output_document.rb
```diff
@@ -58,6 +63,8 @@ def variant
   end
 
   def greeting
```
gjvis: `greeting` doesn't seem appropriate for this text. Perhaps `lead` would be better? (or `lede`)

stevenwilkin: +1 for `lead`

spec/domain/output_document_spec.rb
```diff
@@ -129,8 +129,10 @@ def excludes_all_circumstances
     end
   end
 
-  describe '#greeting' do
-    it 'needs details'
+  describe '#lead' do
+    subject { output_document.lead }
+
+    it { is_expected.to eq('You recently had a Pension Wise guidance appointment with A Guider from The Pensions Advisory Service on March 30, 2015') }
```
andrewgarner: "A Guider" isn't representative test data - it should be a first name.

The date should be formatted as "30 March 2015" as per the GOV.UK style guide.

stevenwilkin: The date format will be addressed with #81

I'll update the guider name

app/templates/output_document_ineligible.html.erb
```diff
@@ -11,12 +11,7 @@
   <p>Dear <%= attendee_name %>
     ,</p>
 
-  <p class="emphasize">You recently had a Pension Wise guidance appointment
-    with <%= guider_name %> from
-
-    <%= guider_organisation %>
-
-    on <%= appointment_date %>.</p>
+  <p class="emphasize"><%= lead %>.</p>
```
andrewgarner: This line doesn't have a punctuation mark in the artwork. We should remove it.

stevenwilkin: Sorted

PR #86, opened 2015-03-30
Format date consistent with the GOV.UK service manual

stevenwilkin:
Addresses #81

### DIFF
--- a/app/domain/output_document.rb
+++ b/app/domain/output_document.rb
@@ -21,7 +21,7 @@ class OutputDocument
   end
 
   def appointment_date
-    appointment_summary.date_of_appointment.to_s(:long)
+    appointment_summary.date_of_appointment.to_s(:gov_uk)
   end
 
   def guider_organisation
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -1,0 +1,3 @@
+# https://www.gov.uk/service-manual/user-centred-design/resources/patterns/dates.html#formatting-dateso
+
+Date::DATE_FORMATS[:gov_uk] = '%e %B %Y'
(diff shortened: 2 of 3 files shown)


app/domain/output_document.rb
```diff
@@ -21,7 +21,7 @@ def attendee_name
   end
 
   def appointment_date
-    appointment_summary.date_of_appointment.to_s(:long)
+    appointment_summary.date_of_appointment.strftime('%e %B %Y')
```
gjvis: I think its worth extracting this as a constant, and then adding a comment to its declaration that calls out why we're using this format, and links to the relevant bit of the GOV.UK service manual.

stevenwilkin: Any suggestions as to where the constant should be declared?

stevenwilkin: I'll declare it in the domain object for now, it can be moved if we need to generate dates in other places later down the line

andrewgarner: Maybe http://guides.rubyonrails.org/i18n.html#adding-date-time-formats






On Monday, 30 Mar 2015 at 17:12, Steven Wilkin <notifications@github.com>, wrote:
Message contents can't be shown right now

stevenwilkin: @gjvis has pointed out Rails' [Date#to_s](http://api.rubyonrails.org/classes/Date.html#method-i-to_formatted_s) which looks like a winner

stevenwilkin:
Updated with date format defined in an initialiser

gjvis:
Better I think, @andrewgarner ?

andrewgarner:
:+1:

PR #87, opened 2015-03-31
Output document stylesheet ERB

stevenwilkin:


gjvis:
If we make the `<style>` tags the responsibility of this method, then we're free to rather inject a `<link>` tag instead without changing the templates. I think this will make refactoring the `OutputDocument` to isolate rendering responsibilities easier.

stevenwilkin:
Good call @gjvis

stevenwilkin:
Done

gjvis:
:+1: :shipit:

PR #88, opened 2015-03-31
Add placeholder for cover letter in output document

stevenwilkin:
Arguably the cover letter could be part of the introduction page, any thoughts?


gjvis:
I think this looks good. @aduggin?

aduggin:
:+1:

PR #92, opened 2015-04-01
Align `OutputDocument` fields with data dictionary

stevenwilkin:
Per my comment on the doc, the data dictionary mentions `guider_first_name` but I believe the printhouse are expecting `guider_name` which we already have @andrewgarner ?

### DIFF
--- a/app/domain/output_document.rb
+++ b/app/domain/output_document.rb
@@ -66,6 +66,10 @@ class OutputDocument
     "#{appointment_summary.id}/#{appointment_summary.reference_number}"
   end
 
+  def format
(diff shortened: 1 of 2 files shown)


andrewgarner:
The print house are working to the data dictionary. It would make sense for us to do so also.

gjvis:
The sample data we gave the print house doesn't match the data dictionary.

stevenwilkin:
Per discussion, `guider_first_name` it is. I'll add another commit to this PR.

stevenwilkin:
bump

andrewgarner:
:+1:

PR #93, opened 2015-04-01
[WIP] Extract HTML generation into a separate renderer

stevenwilkin:
Would appreciate any thoughts on this approach.

Things are confused by the use of `ERB`. `ERB#result` expects an instance of `Binding` and in our case I want to combine the context of the `OutputDocument` along with the context of the rendering object so as to make `stylesheet` available to the template.

To go down this path I've chosen to use `SimpleDelegator` in the renderer and this results in the confusing situation of having the `HTMLRenderer` delegate to the `OutputDocument` which then delegates to the `AppointmentSummary`.

This also means that the specs need to use full instances of `OutputDocument` *and* `AppointmentSummary` instead of doubles :(


gjvis:
Help is at hand: https://gist.github.com/gjvis/cda554a8a10aaade0022

gjvis:
So perhaps extract a rendering context with the delegation stuff, stylesheets, any helpers etc.

But that might end up being class sprawl, which might also feel horrible.

I don't think there's anything :( about your approach - why do the specs need full object and not mocks?

gjvis:
Yeah, looking over what you've done I think your approach is good. My gist doesn't really help much (although if we do want to use collaborators as contexts at least we know how).

In terms of the tests: if we pass in the path to the templates then we could inject test templates that don't depend on any of the delegation stuff, and are easier to verify. (the gherkins will still catch errors in the real templates). Could get rid of a load of the attribute noise in the specs as well.

:+1:

app/domain/output_document/html_renderer.rb
```diff
@@ -0,0 +1,51 @@
+class OutputDocument
+  class HTMLRenderer < SimpleDelegator
+    attr_accessor :output_document
```
andrewgarner: Why is `output_document` exposed publicly?

app/domain/output_document/html_renderer.rb
```diff
@@ -0,0 +1,51 @@
...
+    def pages_to_render
+      case output_document.variant
+      when 'tailored'
+        circumstances = %i(continue_working unsure leave_inheritance
+                           wants_flexibility wants_security wants_lump_sum
+                           poor_health).select do |c|
+          output_document.public_send(c)
+        end
+
+        [:cover_letter, :introduction, circumstances, :other_information].flatten
```
andrewgarner: `%i(cover_letter introduction circumstances other_information)`

Thats the style we follow above and below.

gjvis: That style won't work as `circumstances` is a reference to a local variable

app/domain/output_document/html_renderer.rb
```diff
@@ -0,0 +1,51 @@
+class OutputDocument
+  class HTMLRenderer < SimpleDelegator
+    attr_accessor :output_document
+
+    def initialize(output_document)
+      @output_document = output_document
+      __setobj__(output_document)
+    end
+
+    def pages_to_render
```
gjvis: Similar to @andrewgarner's question above - should this be exposed publicly like this?

stevenwilkin:
Updated with another approach that exposes all values to the template through an `output_document` accessor. This should also address your concern about the visibility of it @andrewgarner

app/domain/output_document/html_renderer.rb
stevenwilkin: I drove this out with specs before writing the template rendering method

PR #94, opened 2015-04-02
Add a CSV renderer

stevenwilkin:
Is this congruent with the test CSV you've generated for the print house @andrewgarner ?


app/domain/output_document/csv_renderer.rb
```diff
@@ -0,0 +1,25 @@
...
+                value_of_pension_pots income_in_retirement continue_working
+                unsure leave_inheritance wants_flexibility wants_security
+                wants_lump_sum poor_health)
+
+    def initialize(output_document)
+      @output_document = output_document
+    end
+
+    def render
+      CSV.generate(col_sep: '|') do |csv|
```
andrewgarner: These are the options I used:
```rb
{ headers: true, col_sep: '|', converters: nil, encoding: 'iso-8859-1:UTF-8' }
```

stevenwilkin: Cool, I'll add in the other options, apart from `headers` as we're only interested in that when we're batching up 1 or more `OutputDocument`s for upload to the print house

stevenwilkin: Looks like `converters` is `nil` by default:

http://ruby-doc.org/stdlib-1.9.2/libdoc/csv/rdoc/CSV.html#DEFAULT_OPTIONS

andrewgarner: :+1:

stevenwilkin: Do you mean `encoding` to be `iso-8859-1`? Ruby complains about `iso-8859-1:UTF-8`:
```
ArgumentError:
       unknown encoding name - iso-8859-1:UTF-8
```

PR #98, opened 2015-04-07
CSV renderer for `OutputDocument`s

stevenwilkin:
A reworking of #73

### DIFF
--- a/app/domain/csv_renderer.rb
+++ b/app/domain/csv_renderer.rb
@@ -1,0 +1,29 @@
+require 'csv'
+
+class CSVRenderer
+  attr_reader :output_documents
+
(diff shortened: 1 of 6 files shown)


gjvis:
I think that this is better than before.

I'm still uneasy about the `OutputDocument#csv` method, I also think that we probbaly should be passing the headers into the `CSVRowRenderer`.

But, without the consuming classes (i.e. the stuff that actually interacts with the print house), its hard to know which way to go, so I think we should merge this in and get started on the print house integration, working outside-in until we hit the point that we need this code and can reflect on its design.

andrewgarner:
:+1:

PR #99, opened 2015-04-08
AppointmentSummary Batching

stevenwilkin:
First attempt at batching up `AppointmentSummary`s to be processed and uploaded to the print house, thoughts?

### DIFF
--- a/app/models/appointment_summary.rb
+++ b/app/models/appointment_summary.rb
@@ -29,6 +29,7 @@ class AppointmentSummary < ActiveRecord::Base
   TITLES = %w(Mr Mrs Miss Ms Mx Dr)
 
   belongs_to :user
+  has_many :appointment_summaries_batches
 
   validates :title, presence: true, inclusion: { in: TITLES, allow_blank: true }
   validates :last_name, presence: true
--- a/app/models/appointment_summary.rb
+++ b/app/models/appointment_summary.rb
@@ -2,6 +2,11 @@ require 'json'
 require 'uri'
 
 class AppointmentSummary < ActiveRecord::Base
+  scope :unprocessed, lambda {
+    includes(:appointment_summaries_batches)
+      .where(appointment_summaries_batches: { appointment_summary_id: nil })
+  }
+
   PostcodeValidator = Class.new(ActiveModel::EachValidator) do
     def validate_each(record, attribute, value)
       return if value.blank?
--- a/app/services/create_batch.rb
+++ b/app/services/create_batch.rb
@@ -1,0 +1,9 @@
+class CreateBatch
+  def call
+    unprocessed_appointment_summaries = AppointmentSummary.unprocessed
+
+    return nil if unprocessed_appointment_summaries.empty?
+
+    Batch.create(appointment_summaries: unprocessed_appointment_summaries)
+  end
+end
--- a/db/migrate/20150408150105_create_batches.rb
+++ b/db/migrate/20150408150105_create_batches.rb
@@ -1,0 +1,9 @@
+class CreateBatches < ActiveRecord::Migration
+  def change
+    create_table :batches do |t|
+      t.datetime :processed_at
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/spec/models/appointment_summary_spec.rb
+++ b/spec/models/appointment_summary_spec.rb
@@ -101,4 +101,36 @@ RSpec.describe AppointmentSummary, type: :model do
       it { is_expected.to be_custom_guidance }
     end
   end
+
+  describe '.unprocessed' do
+    subject { described_class.unprocessed }
+
+    def create_appointment_summary
+      AppointmentSummary.create(
+        title: 'Mr', last_name: 'Bloggs', date_of_appointment: Date.today,
+        reference_number: '123', guider_name: 'Jimmy', guider_organisation: 'tpas',
+        address_line_1: '29 Acacia Road', town: 'Beanotown', postcode: 'BT7 3AP',
+        has_defined_contribution_pension: 'yes', income_in_retirement: 'pension')
+    end
+
+    context 'with no unprocessed items' do
+      let!(:appointment_summaries) { 2.times.map { create_appointment_summary } }
+      let!(:previous_batch) { Batch.create(appointment_summaries: appointment_summaries) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context 'with unprocessed items' do
+      let!(:appointment_summaries) { 2.times.map { create_appointment_summary } }
+
+      it { is_expected.to eq(appointment_summaries) }
+    end
+
+    context 'with processed and unprocessed items' do
+      let!(:appointment_summaries) { 4.times.map { create_appointment_summary } }
+      let!(:previous_batch) { Batch.create(appointment_summaries: appointment_summaries[0..1]) }
+
+      it { is_expected.to eq(appointment_summaries[2, 3]) }
+    end
+  end
 end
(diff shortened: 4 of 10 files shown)


gjvis:
Looks good, although I'd update the tests to verify that multiple appointment summaries are included in the created batches.

andrewgarner:
RuboCop says no.

stevenwilkin:
Just appeasing Rubocop now

app/services/create_batch.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBatch
+  def call
+    appointment_summaries = AppointmentSummary
+                            .joins('LEFT OUTER JOIN appointment_summaries_batches ' \
```
andrewgarner: Can we not achieve this with Active Record without having to resort to half writing SQL?

stevenwilkin: I'll see what I can do, Arel has no concept of an outer join but there may be another way of finding `AppointmentSummary`s that don't have an associated `Batch`

stevenwilkin:
Updated to remove hand-written SQL and to use multiple `AppointmentSummary`s with associated join table entries in the specs.

app/services/create_batch.rb
```diff
@@ -0,0 +1,12 @@
+class CreateBatch
+  def call
+    appointment_summaries = AppointmentSummary
+                            .includes(:appointment_summaries_batches)
+                            .where('appointment_summaries_batches.appointment_summary_id': nil)
+                            .references(:appointment_summaries_batches)
```
gjvis: I think a comment of what this actually does is a good idea. I could understand it more when it was SQL!

stevenwilkin: Turns out we don't need `references`!

app/services/create_batch.rb
```diff
@@ -0,0 +1,11 @@
+class CreateBatch
+  def call
+    appointment_summaries = AppointmentSummary
```
andrewgarner: This could be more expressive. I have to read the AR code to try and work what type of `AppointmentSummary` these are.

We could choose a more descriptive variable name and/or we could scope this maybe?

```rb
class AppointmentSummary
  scope :unprocessed, -> do
    where('appointment_summaries_batches.appointment_summary_id': nil)
      .includes(:appointment_summaries_batches)
  end
end

unprocessed_appointment_summaries = AppointmentSummary.unprocessed
```

db/migrate/20150408150105_create_batches.rb
```diff
@@ -0,0 +1,9 @@
+class CreateBatches < ActiveRecord::Migration
+  def change
+    create_table :batches do |t|
+      t.datetime :processed_at
+
+      t.timestamps null: false
```
andrewgarner: Wouldn't it be useful to know when a `Batch` was created?

stevenwilkin: This results in the same timestamps as [for AppointmentSummary](https://github.com/guidance-guarantee-programme/output/pull/99/files#diff-1acd2e7e27a227829d5d14a91c863bb6L20).

app/services/create_batch.rb
stevenwilkin: Good call, that adds a good bit of clarity

app/services/create_batch.rb
```diff
@@ -0,0 +1,11 @@
...
+                            .includes(:appointment_summaries_batches)
+                            .where('appointment_summaries_batches.appointment_summary_id': nil)
+
+    return nil if appointment_summaries.empty?
+
+    Batch.create(appointment_summaries: appointment_summaries)
```
andrewgarner: If this fails do we expect the caller to know the internals of the `call` method or should the caller provide a failure method to invoke?

stevenwilkin: The intention is that this class will be consumed by *another* class, something along the lines of `ProcessOutputDocuments` which would then coordinate the CSV generation and SFTP upload. Without the consumer in place the interface to this is a bit theoretical.

My assumption was that this call would return a `Batch` only if there were `AppointmentSummary`s which required processing and which were then batched up accordingly.

gjvis: (or throw an exception on failure)

stevenwilkin: Not having anything to process isn't necessarily a failure condition

stevenwilkin:
Rebased, squashed and using a scope on `AppointmentSummary`

spec/models/appointment_summary_spec.rb
```diff
@@ -101,4 +101,36 @@
...
+    def create_appointment_summary
+      AppointmentSummary.create(
+        title: 'Mr', last_name: 'Bloggs', date_of_appointment: Date.today,
+        reference_number: '123', guider_name: 'Jimmy', guider_organisation: 'tpas',
+        address_line_1: '29 Acacia Road', town: 'Beanotown', postcode: 'BT7 3AP',
+        has_defined_contribution_pension: 'yes', income_in_retirement: 'pension')
+    end
+
+    context 'with no unprocessed items' do
+      let!(:appointment_summaries) { [0, 1].map { create_appointment_summary } }
```
gjvis:
```ruby
2.times { create_appointment_summary }
````?

Comment (stevenwilkin) [inline spec/models/appointment_summary_spec.rb]: I tried that originally but it doesn't return the items created

Comment (gjvis) [inline spec/models/appointment_summary_spec.rb]: ah of course!

how about

```ruby
2.times.map { create_appointment_summary }
```

stevenwilkin: I like it!

spec/services/create_batch_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe CreateBatch, '#call' do
+  let(:batch) { Batch.last }
```
gjvis: I don't think that this should be a `let`, it obfuscates the tests a bit

app/models/appointment_summary.rb
```diff
@@ -2,6 +2,11 @@
 require 'uri'
 
 class AppointmentSummary < ActiveRecord::Base
+  scope :unprocessed, lambda {
+    includes(:appointment_summaries_batches)
+      .where('appointment_summaries_batches.appointment_summary_id': nil)
```
andrewgarner: I think you can do this:

```rb
.where(appointment_summaries_batches: { appointment_summary_id: nil })
```

db/migrate/20150408150105_create_batches.rb
andrewgarner: Ignore my stupidity.

app/models/appointment_summary.rb
stevenwilkin: Yep, that works and means not using a string for the condition

andrewgarner: :+1:

stevenwilkin:
Updated to take into consideration all feedback

spec/models/appointment_summary_spec.rb
```diff
@@ -101,4 +101,36 @@
...
+
+    context 'with no unprocessed items' do
+      let!(:appointment_summaries) { 2.times.map { create_appointment_summary } }
+      let!(:previous_batch) { Batch.create(appointment_summaries: appointment_summaries) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context 'with unprocessed items' do
+      let!(:appointment_summaries) { [0, 1].map { create_appointment_summary } }
```
gjvis: `2.times`

gjvis:
do-it

PR #103, opened 2015-04-10
Printhouse SFTP

stevenwilkin:
Need confirmation of paths for both files and if the trigger file needs any specific contents


app/services/upload_to_print_house.rb
```diff
@@ -6,5 +8,17 @@ def initialize(csv)
...
   def call
+    upload_file('/Data.in/batch.csv', csv)
+    upload_file('/Trigger.in/trigger.txt', 'trigger')
+  end
+
+  private
+
+  def upload_file(path, contents)
+    Net::SFTP.start(ENV['SFTP_HOST'], ENV['SFTP_USER'], password: ENV['SFTP_PASSWORD']) do |sftp|
+      sftp.file.open(path, 'w') do |f|
```
gjvis: Need to use `sftp.upload` or the encoding goes screwy

stevenwilkin:
We still need an end-to-end test to confirm the upload takes place.

I've played around with the following Python SFTP server which may be something we want to use:

https://github.com/rspivak/sftpserver

spec/services/upload_to_print_house_spec.rb
```diff
@@ -2,8 +2,23 @@
...
+    allow(uploader).to receive(:upload_file)
+  end
+
+  it 'uploads the CSV' do
+    expect(uploader).to receive(:upload_file).with(csv_path, csv)
+    uploader.call
+  end
+
+  it 'uploads the trigger file' do
+    expect(uploader).to receive(:upload_file).with(trigger_path, trigger)
```
gjvis: I think you can use `have_received` in rspec 3+

gjvis:
:+1:, and we'll sort out an end to end test once we've pulled the features across from the private repo

stevenwilkin:
Updated!

gjvis:
```
     .  o ..                  
     o . o o.o                
          ...oo               
            __[]__            
         __|_o_o_o\__         
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^ 
```

stevenwilkin:
Love it!

app/services/upload_to_print_house.rb
```diff
@@ -6,5 +9,16 @@ def initialize(csv)
   end
 
   def call
+    upload_file('/Data.in/batch.csv', csv)
```
andrewgarner: We know that filename should be:

```rb
"/Data.in/pensionwise_output_#{Date.today.strftime('%Y%m%d')}.csv"
```

stevenwilkin: Merci beaucoup!

app/services/upload_to_print_house.rb
```diff
@@ -6,5 +9,16 @@ def initialize(csv)
...
+    upload_file('/Trigger.in/trigger.txt', 'trigger')
+  end
+
+  private
+
+  def upload_file(path, contents)
+    Net::SFTP.start(ENV['SFTP_HOST'], ENV['SFTP_USER'], password: ENV['SFTP_PASSWORD']) do |sftp|
+      io = StringIO.new(contents)
```
andrewgarner: Minor point, but, you don't need the SFTP connection open to create the `StringIO` instance.

```rb
io = StringIO.new(contents)

Net::SFTP.start(ENV['SFTP_HOST'], ENV['SFTP_USER'], password: ENV['SFTP_PASSWORD']) do |sftp|
  sftp.upload!(io, path)
end
```

gjvis: I'd argue the opposite direction - the `StringIO` isn't needed outside the block so should be scoped to the block only.

gjvis: Actually, I think @andrewgarner is right

andrewgarner: But then you're keeping the connection open so that you can do local processing.

stevenwilkin:
Fixup commits added

gjvis:
Looks good I think. We need to add some logging into this, but that can be a separate PR

stevenwilkin:
Manually merged into master as this was part of a trio of PRs

PR #104, opened 2015-04-13
Add document-contents features

gjvis:
Taken verbatim from the PDF-generation-only repo. I think they're valid as is for this one.


stevenwilkin:
:+1:

PR #106, opened 2015-04-14
Preview record of guidance feature

gjvis:
This PR is bigger than normal, and includes some commented code, as its the first of a set of PRs to pull across the cucumber features and step definitions from the PDFOutput project.


gjvis:
The approach I've taken to validate the preview functionality of the app is to mostly duplicate the "record of guidance contents" feature. Its the least amount of effort given the implementation from the PDFOutput project, and I think is the right trade-off between gherkin readability, step-definition complexity, and test maintainability.

.rubocop.yml
```diff
@@ -20,3 +20,7 @@ Metrics/MethodLength:
 
 Style/RescueModifier:
   Enabled: false
+
+ClassLength:
+  Exclude:
+    - 'features/support/pages/appointment_summary_page.rb'
```
stevenwilkin: We're already excluding a rule from [app/models/appointment_summary.rb](https://github.com/guidance-guarantee-programme/output/blob/bc53223ad624497c1953d55795410d7f9e804e69/app/models/appointment_summary.rb#L88) in the source; do we want to put all excludes in this top-level file or in the individual files themselves?

gjvis: Yeah, you're right - I'll move it to inline

stevenwilkin:
:+1:

PR #110, opened 2015-04-14
Implement "preview ineligibility" feature

gjvis:
Includes commented out code as part of copying across feature implementation from PDFOutput


stevenwilkin:
:+1:

PR #115, opened 2015-04-14
[DO NOT MERGE] Test SFTP upload with a real(ish) server

stevenwilkin:
Simple proof-of-concept demonstrating SFTP upload to a stubbed SFTP server.

The server is a Python package which can be installed with `pip` thus:
```
sudo pip install git+https://github.com/stevenwilkin/sftpserver.git@swallow-eof-error
```

This server [permits all username and password combinations](https://github.com/stevenwilkin/sftpserver/blob/master/src/sftpserver/stub_sftp.py#L33-L35) which is something me might want to think about if we go forward with this approach for our end-to-end tests.


gjvis:
I think that this is worthwhile, although I'd add an assertion to test that trigger file gets created as well.

stevenwilkin:
I've updated this to test that the trigger file exists.

The Python package needs to be installed as root, any preference on how we document this and if it is included in `bin/setup`?

stevenwilkin:
Also, the repo ownership should be transferred away from myself

Issue #108, opened 2015-04-14
Consistent naming: "output document" vs "record of guidance" vs "xxx letter".


gjvis:
We have quite concrete meanings for the terms "output document", "record of guidance", "ineligibility letter" and "covering letter". They are, however, used somewhat inconsistently in the code and features.

We should review the meanings, make sure we all agree on them, and then roll out appropriate changes to make the code and language consistent.

PR #117, opened 2015-04-15
Record of guidance contents feature

gjvis:


features/step_definitions/record_of_guidance_contents_steps.rb
```diff
@@ -22,35 +22,44 @@
...
+  appointment_summary_page.fill_in(@appointment_summary)
+  appointment_summary_page.submit.click
+
+  preview_page = RecordOfGuidancePreviewPage.new
+  preview_page.confirm.click
+
+  ProcessOutputDocuments.new.call
+end
+
+Then(/^the sections it includes should be \(in order\):$/) do |table|
```
gjvis: Perhaps we should remove the "in order", as it's not actually validated (or even controlled by this app) at all?

It does better express intention though.

andrewgarner: I think we should remove. There is a logical order to the document that we don't call out in any of the other steps, e.g. the pension pot details page sits before the options page. It's misleading to give special attention to the order here when the app has no impact on it.

gjvis: Cool, its called out in the preview features (where it is the responsibility of the app), so the intention is not entirely lost.

Will remove.

features/step_definitions/record_of_guidance_contents_steps.rb
```diff
@@ -97,35 +106,45 @@
...
-#   output_document = OutputDocument.new(@appointment_summary)
+Then(/^the record of guidance should include the details of the appointment$/) do
+  output_document = OutputDocument.new(@appointment_summary)
 
-#   expect(page).to have_content(output_document.appointment_date)
-#   expect(page).to have_content(output_document.appointment_reference)
-#   expect(page).to have_content(output_document.guider_first_name)
-#   expect(page).to have_content(output_document.guider_organisation)
-# end
+  expected = {
```
andrewgarner: What about `lead`? It's a string that includes these fields.

features/support/matchers/expect_uploaded_csv_to_include.rb
```diff
@@ -0,0 +1,27 @@
...
+
+CSV::Converters[:boolean] = lambda do |value|
+  case value.to_s
+  when 'true' then true
+  when 'false' then false
+  else value
+  end
+end
+
+module UploadedCsvMatcher
```
andrewgarner: If we're following the GitHub Ruby style guide we should keep acronyms like CSV uppercase.

features/step_definitions/record_of_guidance_contents_steps.rb
gjvis: The lead isn't actually in the record of guidance, only the covering letter, so the assertion is part of that feature.

Somewhat related: I've raised an issue (#108), for us to review the naming of the various documents to make sure its consistent and clear.

features/support/matchers/expect_uploaded_csv_to_include.rb
gjvis: :+1:

gjvis:
Ok, I've cleaned up the commits a bit. Good to go? /ping @stevenwilkin @andrewgarner

PR #118, opened 2015-04-16
Better login for Cucumber

### LINKED ISSUE #112: Remove the need for "create and log in a user" backgrounds to all scenarios
I think it'd be better to handle that with a Before hook, and then have a tag to allow scenarios to bypass the login hook.

gjvis:
Closes #112

### DIFF
--- a/features/support/authentication.rb
+++ b/features/support/authentication.rb
@@ -1,7 +1,6 @@
-Given(/^a new, authenticated user$/) do
+Before('~@unauthenticated') do
   email = 'guider@example.com'
   password = 'pensionwise'
-
   User.create(email: email, password: password, password_confirmation: password).confirm!
 
   page = UserSignInPage.new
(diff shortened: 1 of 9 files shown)


features/support/authentication.rb
```diff
@@ -1,9 +1,11 @@
-Given(/^a new, authenticated user$/) do
-  email = 'guider@example.com'
-  password = 'pensionwise'
+email = 'guider@example.com'
+password = 'pensionwise'
 
+unless User.exists?(email: email)
   User.create(email: email, password: password, password_confirmation: password).confirm!
```
stevenwilkin: Could potentially use [User#first_or_create](http://guides.rubyonrails.org/v3.2.17/active_record_querying.html#first_or_create)?

gjvis: Definitely. Will change it.

stevenwilkin:
Looks good, I like that we're removing the repeated background step in all those features

gjvis:
This now includes an updated RSpec spec as well

PR #121, opened 2015-04-16
Use fixture templates when testing the HTML renderer

stevenwilkin:
The Cucumber features now test the rendered HTML thouroughly so makes sense to reduce the duplication in testing effort


gjvis:
:+1: Those are some good PR stats...

PR #122, opened 2015-04-16
Add and implement accessibility feature

gjvis:


stevenwilkin:
:+1:

gjvis:
This has now been updated with functionality to exclude braille until the print house can handle it, and the method of sending them braille is decided. This is expected to be resolved within the next two weeks.

ping @andrewgarner, @stevenwilkin - look ok?

andrewgarner:
Can we handle large print yet?

gjvis:
Good point, we should exclude both.

app/services/create_batch.rb
```diff
@@ -2,6 +2,11 @@ class CreateBatch
   def call
     unprocessed_appointment_summaries = AppointmentSummary.unprocessed
 
+    # exclude braille until the print house can actually deal with it
+    unprocessed_appointment_summaries = unprocessed_appointment_summaries.reject do |as|
```
andrewgarner: How about chaining the scopes?

```rb
AppointmentSummary.standard_format_preference.unprocessed
```

gjvis: @stevenwilkin and I discussed this a bit, and it would add more moving parts for not much benefit on something which is short-lived. If we start blowing memory limits on dynos than obviously that changes things, and we'd have to revisit this.

gjvis:
I've taken the "do not process unsupported" stuff out of this PR, will raise seperately

PR #123, opened 2015-04-17
Exclude braille and large text from batches

gjvis:
The print house can't currently deal with braille or large text, so we
should not send output documents in those formats to them

### DIFF
--- a/app/services/create_batch.rb
+++ b/app/services/create_batch.rb
@@ -1,6 +1,6 @@
 class CreateBatch
   def call
-    unprocessed_appointment_summaries = AppointmentSummary.unprocessed
+    unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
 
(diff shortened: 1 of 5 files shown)


app/services/create_batch.rb
```diff
@@ -1,6 +1,6 @@
 class CreateBatch
   def call
-    unprocessed_appointment_summaries = AppointmentSummary.unprocessed
+    unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
```
gjvis: Should we pull out a scope here? This class is already tightly coupled to ActiveRecord in any case

gjvis:
BUMP

(in other news, searching for "bump gif" is a bad idea... :hushed: )

stevenwilkin:
thumbs up

If the print house will eventually handle the differing formats for us then this commit will be easy to rollback

PR #124, opened 2015-04-17
Remove "create record of guidance" feature

gjvis:
The functionality that it was validating is better covered by other
features


stevenwilkin:
kill it with fire

PR #129, opened 2015-04-20
Process output documents schedule

stevenwilkin:
The schedule has to be added manually as it can't be configured via `app.json`

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
+process_output_documents: bundle exec rake output_documents:process
 web: bin/puma -C config/puma.rb
 worker: bin/sidekiq -C config/sidekiq.yml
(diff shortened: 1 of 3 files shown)


stevenwilkin:
`Procfile` entry added

gjvis:
:+1:

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bin/puma -C config/puma.rb
 worker: bin/sidekiq -C config/sidekiq.yml
+process_output_documents: rake output_documents:process
```
stevenwilkin: `rake` or `bin/rake` ?

`bin/puma` and `bin/sidekiq` are Bundler shims whereas `bin/rake` [loads Spring](https://github.com/guidance-guarantee-programme/output/blob/master/bin/rake#L3)

stevenwilkin: Also, alphabetical ordering @andrewgarner ?

andrewgarner: Yes please. Likewise for the add ons.

PR #131, opened 2015-04-21
Add logging to output document processing

gjvis:
This adds logging to the output document processing without refactoring or rewriting how the processing itself works.

I'm not super happy with it, but it is the least amount of change to get logging in. I think this area needs attention, after which the logging will (along with everything else) hopefully end up cleaner.

### DIFF
--- a/app/services/create_batch.rb
+++ b/app/services/create_batch.rb
@@ -1,6 +1,24 @@
 class CreateBatch
+  def initialize(logger = Rails.logger)
+    @logger = logger
+  end
+
   def call
+    logger.debug('CreateBatch started')
+
+    create_batch
+  ensure
+    logger.debug('CreatBatch completed')
+  end
+
+  private
+
+  attr_reader :logger
+
+  def create_batch
     unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
+
+    logger.info("Unprocessed appointment summaries: #{unprocessed_appointment_summaries.count}")
 
     return nil if unprocessed_appointment_summaries.empty?
 
--- a/app/services/process_output_documents.rb
+++ b/app/services/process_output_documents.rb
@@ -1,15 +1,58 @@
 class ProcessOutputDocuments
+  def initialize(logger = Rails.logger)
+    @logger = logger
+  end
+
   def call
-    batch = CreateBatch.new.call
+    run_id = SecureRandom.uuid
+    logger.tagged('ProcessOutputDocuments', run_id) do |logger|
+      Runner.new(logger).call
+    end
+  end
 
-    return nil unless batch
+  private
 
-    output_documents = batch.appointment_summaries.map do |appointment_summary|
-      OutputDocument.new(appointment_summary)
+  attr_reader :logger
+
+  class Runner
+    attr_reader :logger
+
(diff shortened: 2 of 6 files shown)


app/services/upload_to_print_house.rb
```diff
@@ -15,10 +16,27 @@ def call
...
-    Net::SFTP.start(ENV['SFTP_HOST'], ENV['SFTP_USER'], password: ENV['SFTP_PASSWORD']) do |sftp|
+    Net::SFTP.start(host, user, password: password) do |sftp|
+      logger.info("Connected to #{host}")
+
       sftp.upload!(io, path)
+      logger.info("#{path} uploaded")
     end
   end
+
+  def host
```
stevenwilkin: Moving the various params into methods would be best in a separate commit

app/services/create_batch.rb
```diff
@@ -1,9 +1,19 @@
 class CreateBatch
+  def initialize(logger = Rails.logger)
+    @logger = logger
+  end
+
   def call
     unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
 
+    logger.info("Unprocessed appointment summaries: #{batch.appointment_summaries.count}")
```
andrewgarner: Where does `batch` come from?

app/services/upload_to_print_house.rb
gjvis: Good point.

app/services/create_batch.rb
```diff
@@ -1,9 +1,19 @@
 class CreateBatch
+  def initialize(logger = Rails.logger)
+    @logger = logger
+  end
+
   def call
     unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
```
stevenwilkin: Maybe a logging statement to lets us know this has been kicked off?

app/services/create_batch.rb
gjvis: Last-minute-change oversight :(

app/services/process_output_documents.rb
```diff
@@ -1,6 +1,15 @@
 class ProcessOutputDocuments
   def call
-    batch = CreateBatch.new.call
+    tags = []
+    tags << 'ProcessOutputDocuments'
+    tags << SecureRandom.uuid
```
stevenwilkin:
```ruby
tags = ['ProcessOutputDocuments', SecureRandom.uuid]
```
?

andrewgarner: What's the intent behind `SecureRandom.uuid`? If it is to signify a unique invocation of this method?

gjvis: I'll extract this into another class, and make sure the UUID is named.

app/services/process_output_documents.rb
```diff
@@ -10,6 +19,22 @@ def call
 
     csv = CSVRenderer.new(output_documents).render
 
-    UploadToPrintHouse.new(csv).call
+    UploadToPrintHouse.new(csv, logger).call
+  end
+
+  def call_logged(logger)
```
stevenwilkin: bad taste in the mouth

gjvis:
superceded by #154

PR #135, opened 2015-04-22
Restrict address fields to 50 characters

stevenwilkin:


app/models/appointment_summary.rb
```diff
@@ -32,7 +32,8 @@ def postcode=(str)
   has_many :appointment_summaries_batches, dependent: :destroy
 
   validates :title, presence: true, inclusion: { in: TITLES, allow_blank: true }
-  validates :last_name, presence: true
+  validates :first_name, length: { maximum: 50 }
+  validates :last_name, presence: true, length: { maximum: 50 }
```
andrewgarner: I think the desired behaviour is that the combined `title`, `first_name` and `last_name` can't be greater than 50 characters. This is so the full name can be seen when shown through a C5 envelope window.

app/models/appointment_summary.rb
```diff
@@ -53,9 +54,12 @@ def postcode=(str)
...
 
-  validates :address_line_1, presence: true
-  validates :town, presence: true
-  validates :postcode, presence: true, postcode: true
+  validates :address_line_1, presence: true, length: { maximum: 50 }
+  validates :address_line_2, length: { maximum: 50 }
+  validates :address_line_3, length: { maximum: 50 }
+  validates :town, presence: true, length: { maximum: 50 }
+  validates :county, length: { maximum: 50 }
+  validates :postcode, presence: true, postcode: true, length: { maximum: 50 }
```
andrewgarner: We don't need to validate length of postcode. The postcode validation will handle this.

app/models/appointment_summary.rb
stevenwilkin: Ok!

gjvis: Yeah, its `OutputDocument#attendee_name` that can't be more than 50. Or perhaps more specifically, no individual line of `OutputDocument#attendee_address`

app/models/appointment_summary.rb
stevenwilkin: Good spot, copy'n'paste fail

gjvis:
:+1:

PR #138, opened 2015-04-22
International address support

gjvis:


app/models/appointment_summary.rb
```diff
@@ -58,7 +58,8 @@ def postcode=(str)
   validates :address_line_3, length: { maximum: 50 }
   validates :town, presence: true, length: { maximum: 50 }
   validates :county, length: { maximum: 50 }
-  validates :postcode, presence: true, postcode: true
+  validates :country, inclusion: { in: Countries.all, allow_blank: true, allow_nil: true }
```
andrewgarner: Why do we want to allow the country to be blank?

gjvis: Good point, I'll update.

spec/models/countries_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+require_relative File.join(File.dirname(__FILE__), '..', '..', 'app', 'models', 'countries')
```
stevenwilkin: You can replace these 2 lines with `require 'rails_helper'`.

`spec_helper` [is already required](https://github.com/guidance-guarantee-programme/output/blob/master/.rspec#L2), perhaps we should add `rails_helper` to `.rspec` and remove the require from all the spec files

gjvis: I didn't realise spec_helper was already required.

Required the full rails environment is a ~3000% slowdown loading this file alone, but obviously makes no difference for a full run. I'll change it for consistency, but a better solution would be removing the require rails_helper lines from any specs that don't actually depend on rails.

gjvis:
Ok, country is no longer optional, and I've cleaned up the things from the other comments.

More :eyes: please

app/models/countries.rb
```diff
@@ -0,0 +1,247 @@
+# This list is taken from https://www.gov.uk/api/foreign-travel-advice.json
+# $ curl https://www.gov.uk/api/foreign-travel-advice.json | jq '.details.countries[].name'
+
+module Countries
+  def self.all
```
andrewgarner: It's not clear to me why we want the constant to be private?

Why wouldn't I want to do something like this?

```rb

UK = 'United Kingdom'
ALL = ['Afghanistan', 'Albania', UK, 'Zimbabwe']

```

gjvis: I prefer bare-words to direct constant access, scope is more predictable, and it is simpler to extend.

I made the constant private to make it unambiguous that the methods should be used.

And I didn't use freeze because freezing an Array doesn't freeze its elements, so using it here would be misleading.

app/domain/output_document.rb
```diff
@@ -20,13 +20,16 @@ def attendee_name
   end
 
   def attendee_address
+    attendee_country = Countries.uk?(self.attendee_country) ? nil : self.attendee_country
```
stevenwilkin: Instead of doing this we could perhaps not delegate `attendee_country` to `appointment_summary` and instead add an `attendee_country` method to `OutputDocument` which encapsulates this logic

gjvis: It is a bit weird, but I thought it'd be weirder to have an `attendee_country` method that returns nil.

As discussed, we should make the individual address methods private as a separate PR.

stevenwilkin:
:+1:

PR #139, opened 2015-04-22
[DO NOT MERGE] Do not process non-uk appointment summaries

gjvis:
Dependant on #138 

Comment welcome!


app/services/create_batch.rb
```diff
@@ -1,6 +1,8 @@
 class CreateBatch
   def call
-    unprocessed_appointment_summaries = AppointmentSummary.unprocessed.where(format_preference: :standard)
+    unprocessed_appointment_summaries = AppointmentSummary.unprocessed
+                                        .where(format_preference: :standard)
+                                        .where(country: [nil, '', 'United Kingdom'])
```
gjvis: Think its time to extract these into scopes. We're going to need it for manual processing of these in any case. Perhaps a separate PR though?

stevenwilkin: Yeah, here we've a scope with 2 chained `where`s, definitely a cause for concern

PR #141, opened 2015-04-24
Do not process non-uk appointment summaries

gjvis:
Includes chained `where` statements on top of a scope, which will be resolved in an upcoming PR.


stevenwilkin:
make it so

PR #142, opened 2015-04-24
Allow logged-in users to change their passwords

stevenwilkin:


config/routes.rb
```diff
@@ -23,6 +23,10 @@
   mount GovukAdminTemplate::Engine, at: '/style-guide'
 
   devise_for :users
+  as :user do
+    get 'users/edit' => 'devise/registrations#edit', as: 'edit_user_registration'
+    put 'users/:id' => 'devise/registrations#update', as: 'user_registration'
+  end
```
stevenwilkin: Could tidy this up like this:
```ruby
  as :user do
    scope path: 'users', controller: 'devise/registrations' do
      get 'edit', action: 'edit', as: 'edit_user_registration'
      put ':id', action: 'update', as: 'user_registration'
    end 
  end
```

gjvis:
looks good :ship:

stevenwilkin:
For completeness. It adds this link:

image

Which leads to this form:

image

stevenwilkin:
Fixed vertical spacing, pesky `<br/>`s

app/helpers/application_helper.rb
```diff
@@ -13,4 +13,17 @@ def bootstrap_alert_class_for(flash_type)
...
+    return unless model.errors.any?
+
+    content_tag(:div, class: %w(alert alert-danger), role: %w(alert)) do
+      concat(content_tag(:h2, "There's a problem", class: %w(alert__heading)))
+      concat(content_tag(:ul) do
+        model.errors.full_messages.each do |message|
+          concat(content_tag(:li, message))
+        end
+      end)
+    end
```
gjvis: I love the helper, but I think it'd be better if the markup were in a partial...

stevenwilkin:
Voila:

image

app/helpers/application_helper.rb
stevenwilkin: I like the succinctness of invoking it with `error_messages_for(..)` instead of `render partial: 'errors', model: the_model`. The nested `content_tags` are nasty though.

Any opinion @andrewgarner ?

gjvis: I actually meant rendering a partial from the helper. So the API is the same, its just the method of generating the output that changes.

stevenwilkin:
Hmm, should have the bootstrap-style submit button I think...

app/helpers/application_helper.rb
stevenwilkin: Ah, that would work!

stevenwilkin: I'll pour one out for the time spent getting those nested `content_tag`s to work :hankey: :gun:

stevenwilkin:
Like this:

image

stevenwilkin:
Rebased and ready to go

gjvis:
:shipit: 

The guides are going to love us

PR #143, opened 2015-04-27
Self service password reset

stevenwilkin:
It adds this link to the login page:

log in 2015-04-27 11-11-40

Which leads on to:

gov uk 2015-04-27 11-12-06

And finally:

gov uk 2015-04-27 11-12-40

The forms for these last 2 pages are the ones that ship with Devise. Do we care enough to rewrite them to be congruent with our styleguide? If so, do we want to rework the login form?

Also, are we ok with deploying the code change along with the database change or do we want to split this up into 2 deployments?

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,6 @@
 class User < ActiveRecord::Base
   devise :database_authenticatable, :confirmable, :invitable, :lockable, :timeoutable, :trackable,
-         :validatable, :zxcvbnable
+         :validatable, :zxcvbnable, :recoverable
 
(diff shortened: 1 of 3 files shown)


gjvis:
I think we should update so that the spacing is consistent with the login form, its a bit confusing otherwise.

And we can't deploy to this app in a way that incurs downtime, so I think it'll need to be two PRs

stevenwilkin:
Now with updated forms:
image

image

gjvis:
So shiny...
shiny

gjvis:
:shipit:

PR #144, opened 2015-04-27
Upload trigger files on successful upload of data files to print house

gjvis:
`CSVUploadJob` is a weird name, its a class that encapsulates the logic of turning a `Batch` into everything needed to send to the print house (i.e. file contents, file paths). Can anyone think of a better name?

### DIFF
--- a/app/models/csv_upload_job.rb
+++ b/app/models/csv_upload_job.rb
@@ -14,7 +14,21 @@ class CSVUploadJob
   end
 
   def payload_path
+    upload_path('csv')
+  end
+
+  def trigger
+    ''
+  end
+
+  def trigger_path
+    upload_path('trg')
+  end
+
+  private
+
+  def upload_path(extension)
     timestamp = batch.created_at.strftime('%Y%m%d%H%M%S%L')
-    "/Data.in/pensionwise_output_#{timestamp}.csv"
+    "/Data.in/pensionwise_output_#{timestamp}.#{extension}"
   end
 end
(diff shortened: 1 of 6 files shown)


gjvis:
Ok, apart from the weird naming (which might be fine?) I think this is ready for review.

app/models/csv_upload_job.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def trigger_path
+    upload_path('trg')
+  end
+
+  private
+
+  def upload_path(extension)
+    timestamp = batch.created_at.strftime('%Y%m%d%H%M%S%L')
+    "/data.in/pensionwise_output_#{timestamp}.#{extension}"
```
andrewgarner: Is this case sensitive? If so it should be `/Data.in`

gjvis: It well may be, nice catch.

app/models/csv_upload_job.rb
```diff
@@ -0,0 +1,34 @@
...
+  def payload_path
+    upload_path('csv')
+  end
+
+  def trigger
+    ''
+  end
+
+  def trigger_path
+    upload_path('trg')
```
andrewgarner: The current UAT environment has `/Trigger.in` - does that need to change?

gjvis: That directory exists as part of their "standard" SFTP setup. For this integration they're expecting trigger files alongside the data files.

andrewgarner: :+1:

andrewgarner:
:+1:

PR #145, opened 2015-04-27
Add Devise recoverable columns to users table

stevenwilkin:
Migration pulled out of [output#143](https://github.com/guidance-guarantee-programme/output/pull/143)

### DIFF
--- a/db/migrate/20150427100026_add_devise_recoverable_to_users.rb
+++ b/db/migrate/20150427100026_add_devise_recoverable_to_users.rb
@@ -1,0 +1,7 @@
+class AddDeviseRecoverableToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :reset_password_token, :string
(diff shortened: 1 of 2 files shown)


db/migrate/20150427100026_add_devise_recoverable_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeviseRecoverableToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :reset_password_token, :string
```
andrewgarner: Should there be an index? I notice there is one for the confirmation token in the schema.

stevenwilkin: Good spot!

stevenwilkin:
Updated to include index. Laziness will be the death of me!

andrewgarner:
:+1:

PR #146, opened 2015-04-27
Extract error message generation into a helper

stevenwilkin:
Going to need this for the Devise forms that will be pulled in soon

### DIFF
--- a/app/views/shared/_errors.html.erb
+++ b/app/views/shared/_errors.html.erb
@@ -1,0 +1,10 @@
+<% if local_assigns.include?(:model) && local_assigns[:model].errors.any? %>
+  <div class="alert alert-danger" role="alert">
+    <h2 class="alert__heading">There's a problem</h2>
+    <ul>
+      <% local_assigns[:model].errors.full_messages.each do |msg| %>
+        <li><%= msg %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -13,4 +13,8 @@ module ApplicationHelper
       flash_type.to_s
     end
   end
+
+  def error_messages_for(model)
+    render partial: 'shared/errors', locals: { model: model }
+  end
 end
--- a/app/views/appointment_summaries/new.html.erb
+++ b/app/views/appointment_summaries/new.html.erb
@@ -11,16 +11,7 @@
   </div>
 
   <div class="display_if_consent">
-    <% if @appointment_summary.errors.any? %>
-      <div class="alert alert-danger" role="alert">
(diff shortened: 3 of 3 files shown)


stevenwilkin:
Updated with a helper/partial double punch per @gjvis's suggestion

app/views/shared/_errors.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if local_assigns[:model].errors.any? %>
+  <div class="alert alert-danger" role="alert">
+    <h2 class="alert__heading">There's a problem</h2>
+    <ul>
+      <% local_assigns[:model].errors.full_messages.each do |msg| %>
```
gjvis: This call and the one one line 1 will error out with an unuseful error message if model is nil.

app/views/shared/_errors.html.erb
```diff
@@ -0,0 +1,10 @@
...
+        <li><%= msg %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
```
gjvis: The partial actually only needs the errors, not the whole model (although that might be an implementation detail, your call)

app/views/shared/_errors.html.erb
stevenwilkin: That did cross my mind. The helper that renders this partial should never be called if the model is nil but some defensive programming here could be in order

app/views/shared/_errors.html.erb
stevenwilkin: I'm happy to have the model here considering how tightly coupled to `ActiveModel` the whole thing is

stevenwilkin:
Updated with some defensive programming

app/views/shared/_errors.html.erb
gjvis: :+1:

app/views/shared/_errors.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if local_assigns.fetch(:model, nil) && local_assigns[:model].errors.any? %>
```
gjvis: `local_assigns.include?(:model)`

stevenwilkin: Good call, I didn't know that existed, looks much neater

gjvis: I think local_assigns is just a `Hash` (although, knowing Rails, could be anything, hehe)

stevenwilkin:
Build is passing, wanna hit the merge button?

gjvis:
You go for it, I'm eyeballs deep in pub/sub. SO CLOSE TO A PR.

gjvis:
:ship: :ship: :ship:

stevenwilkin:
10 4 good buddy

PR #149, opened 2015-04-28
Upload all unprocessed batches

gjvis:
`ProcessOutputDocuments#call` will now upload all unprocessed (i.e. not yet uploaded) batches.

Adds [Wisper](https://github.com/krisleech/wisper) for pub/sub, which provides far more control than [ActiveSupport::Notifications](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) and better expresses intention.

I think that this can be merged as a single PR, as no production code currently creates or reads batches.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'sidekiq'
 gem 'sinatra', require: nil # Sidekiq UI
 gem 'uglifier', '>= 1.3.0'
+gem 'wisper'
 
 group :development, :test do
   gem 'launchy'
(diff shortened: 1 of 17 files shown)


gjvis:
More :eyes: please, I think this is ready to go. Note the last commit!

stevenwilkin:
Any reason why you don't want `ProcessOutputDocuments` to know about empty `Batch`s?

gjvis:
Now the we aren't passing anything from `CreateBatch#call` into `UploadToPrintHouse` it seemed wrong to have the noop bail-out at that point. `UploadToPrintHouse` will now handle whatever is thrown at it.

I think I've started thinking of both CreateBatch and UploadToPrintHouse as sort-of public API type classes, which influenced this thinking. That's not really true currently though, so I might be off the mark on this one.

Lets hold off on merging this until tomorrow when I'm back in the office.

gjvis:
*points at #151

PR #151, opened 2015-04-28
Upload all unprocessed batches (option B)

gjvis:
Simpler version of #149.

Changes less about the existing implementation, doesn't need pub/sub, fewer moving parts, fewer interdependencies.


gjvis:
I reckon this PR wins on stats alone...

stevenwilkin:
yes

PR #152, opened 2015-04-29
Refactor upload_to_print_house_spec

gjvis:
- asserts upload order
- mocks Net::SFTP rather than a private method on our class. Net::SFTP is the de facto standard for SFTP in Ruby, and has a stable API, so i think that mocking it is acceptable in this case given the ease of refactoring that this allows


stevenwilkin:
ship it

PR #153, opened 2015-04-29
Retry print-house uploads up to 5 times

gjvis:
Adds naïve fault tolerance to the print-house uploads. A more sophisticated implementation would add random exponential backoff, but I'm not sure its needed in this case. Thoughts?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'pg'
 gem 'puma'
 gem 'rails', '4.2.0'
 gem 'rails-i18n', '~> 4.0.0'
+gem 'retriable'
 gem 'sass-rails', '~> 5.0'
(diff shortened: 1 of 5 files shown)


gjvis:
Could add another gem (https://github.com/kamui/retriable) that looks to provide more sophisticated rety logic. It doesn't have any runtime dependancies, and seems pretty small and focussed...

gjvis:
Ok, I've added it - I think its worth it considering the small overhead. Compare the commits for the two options and let me know what you think.

gjvis:
Rebased now that logging is merged in, and added logging for errors when retrying.

@andrewgarner, @stevenwilkin :eyes: please

andrewgarner:
:+1:

PR #154, opened 2015-04-29
Use tagged logging for key stages of PrintHouse integration

stevenwilkin:
Instead of performing everything with a block passed to `ActiveSupport::TaggedLogging#tagged` we can make use of the public API on offer and simply push and pop tags as we see fit. For better or for worse tags are [stored on Thread.current](https://github.com/rails/rails/blob/f046e8a70fe526ebcea88ea1d084eb4b634b653b/activesupport/lib/active_support/tagged_logging.rb#L46). It's up for debate wether this is a valid approach or not.

Produces the following output:

image

Apart from the SQL logging, the key stages logged are:

* start
* id and number of contents upon batch creation
* id of batch to be uploaded
* paths of individual files to be uploaded
* end

A simplification could be `CreateBatch` and `UploadToPrintHouse` accessing `Rails.logger` directly rather than setting an instance variable.

Opinions gents?

### DIFF
--- a/app/services/process_output_documents.rb
+++ b/app/services/process_output_documents.rb
@@ -1,7 +1,20 @@
 class ProcessOutputDocuments
+  attr_reader :tags
+
+  def initialize
+    @tags = [self.class, SecureRandom.uuid]
+  end
+
   def call
+    logger.push_tags(tags)
+    logger.info('Start')
+
     CreateBatch.new.call
     upload_all(Batch.unprocessed)
+
+    logger.info('End')
+  ensure
+    logger.pop_tags(tags.size)
   end
 
   private
@@ -11,8 +24,14 @@ class ProcessOutputDocuments
   end
 
   def upload(batch)
+    logger.info("Processing batch #{batch.id}")
+
     job = CSVUploadJob.new(batch)
     UploadToPrintHouse.new(job).call
     batch.mark_as_uploaded
+  end
+
+  def logger
+    Rails.logger
   end
 end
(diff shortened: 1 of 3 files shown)


app/services/process_output_documents.rb
```diff
@@ -1,7 +1,21 @@
 class ProcessOutputDocuments
+  attr_accessor :logger, :tags
```
gjvis: `attr_reader` might be more appropriate?

gjvis:
On the whole this is solid, nice one.

My only thought is that we should either:
 - have `attr_reader :logger` backed by `@logger` on each class, but then actually pass the logger from `ProcessOutputDocuments` to the other classes, or
 - go for `def logger; Rails.logger end` in each class (i.e. no instance variable, no attr_reader), and just not bother with the initializer params

Having the initializer params but not actually using them anywhere (not even in tests) is a bit messy, and might cause future confusion?

app/services/process_output_documents.rb
stevenwilkin: Good idea

stevenwilkin:
I'm thinking maybe `def logger ...` is the way to go. I originally was going to pass the logger around between objects but when I saw that the tags were stored on `Thread.current` I didn't see the point

gjvis:
Totally, we're already in bed with Rails.

stevenwilkin:
That's decided then. You're totally right about having the initialiser params and not even specifying them in the specs. When in Rails....

gjvis:
deal-with-it-dhh

gjvis:
(I wanted a DHH YOLO gif, but I'm too lazy to make one, and that one sort of fits the bill...)

stevenwilkin:
Updated in relation to feedback

gjvis:
:ship:

PR #156, opened 2015-05-06
Handbuilt artisan csv

stevenwilkin:
The PrintHouse don't understand CSV properly so remove the use of the standard library CSV renderer and generate it by hand


gjvis:
Rubocop issues, otherwise :+1:

PR #157, opened 2015-05-07
Add monitoring to process_output_documents scheduled jobs

gjvis:
Alert [Dead Man's Snitch](https://deadmanssnitch.com/) when output documents are processed


stevenwilkin:
:+1: Just need to set the env var for all the apps

PR #158, opened 2015-05-11
User name

stevenwilkin:
Depend on the `first_name` and `last_name` attributes, ignoring the redundant `name`.

This needs to be run before the deployment/migration to accommodate legacy data:
```ruby
User.where(first_name: nil, last_name: nil).where.not(name: nil).each { |u| u.first_name = u.first_name; u.last_name = u.last_name; u.save }
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,16 +11,4 @@ class User < ActiveRecord::Base
   def name
     [first_name, last_name].join(' ').strip
   end
-
-  def first_name
-    super || self[:name].try do |name|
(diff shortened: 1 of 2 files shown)


gjvis:
I don't think the `name` and `name=` methods are used at all anymore?

stevenwilkin:
Good catch

stevenwilkin:
Updated to remove `User#name=` but keeping `User#name` as it's still used in the layout.

gjvis:
:+1:

PR #160, opened 2015-05-11
[FOR DISCUSSION - DO NOT MERGE] De-duplicate appointment summaries on create

gjvis:
I'm not 100% sure that this is a good idea...


gjvis:
@stevenwilkin can you weigh in on this discussion?

stevenwilkin:
This would help deal with double submissions but would it be any better to disable the submit button with JavaScript instead?

Also, if we were go ahead with this would it be an idea to do some logging around it we could analyse later?

gjvis:
Perhaps all of the above? The solutions are all imperfect :(

We can also deduplicate on *creation* of batches, but it requires quite a bit of reworking of `CreateBatch` =/

stevenwilkin:
So how many of the duplicates you've looked at @gjvis been due to double-submission and how many had other causes?

gjvis:
Over a 2 week period there were about 30 dups. 4 were double submits, the rest were people mis-typing the reference number.

stevenwilkin:
Was the reference number the only field you accounted for @gjvis ? What about the likes of the attendee name and address?

PR #161, opened 2015-05-11
[DO NOT MERGE] Spiked confirmation page of collected details

stevenwilkin:
Renders a table of supplied details before showing the HTML preview:

image


gjvis:
I think this is a good outcome, although we need to make sure we don't "surprise" the users with it. We should take this to the user researches perhaps?

(There's also a fair amount of scope to streamline the code, but I'm assuming that as this is POC we're not commenting on the code itself)

PR #164, opened 2015-05-15
Use SSH for github gems

gjvis:


Gemfile
```diff
@@ -2,6 +2,9 @@ source 'https://rubygems.org'
 
 ruby '2.2.2'
 
+# force Bundler to use HTTPS for github repos
```
stevenwilkin: This forces SSH rather than the git protocol, no?

andrewgarner:
:+1:

Gemfile
andrewgarner: Good spot

andrewgarner:
When the comment is correct, then :+1:

Gemfile
gjvis: Yep, copied and pasted from another repo (with slightly different constraints) in haste :(

PR #165, opened 2015-05-15
Naively prevent double submission of form

stevenwilkin:
Too crude? Any browser compatibility problems spring to mind @aduggin ?

### DIFF
--- a/app/views/appointment_summaries/_preview_form.html.erb
+++ b/app/views/appointment_summaries/_preview_form.html.erb
@@ -7,6 +7,7 @@
   <% end %>
 
   <%= f.submit local_assigns.fetch(:submit_button_text),
-    class: (%w(btn btn-lg) | Array(local_assigns[:submit_button_classes])) %>
+    class: (%w(btn btn-lg) | Array(local_assigns[:submit_button_classes])),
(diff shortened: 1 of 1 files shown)


stevenwilkin:
A potentially quick win to catch ~13% of duplicate records

app/views/appointment_summaries/_preview_form.html.erb
```diff
@@ -7,6 +7,7 @@
   <% end %>
 
   <%= f.submit local_assigns.fetch(:submit_button_text),
-    class: (%w(btn btn-lg) | Array(local_assigns[:submit_button_classes])) %>
+    class: (%w(btn btn-lg) | Array(local_assigns[:submit_button_classes])),
+    onclick: 'this.disabled = true; this.parentElement.submit();' %>
```
gjvis: Good in principle, but makes too many assumptions about the DOM. A markup change could easily break this and automated tests wouldn't catch it.

stevenwilkin: Yeah, I was thinking of where to put the code and this was the simplest possible thing that would work. Hence my apprehension.

PR #166, opened 2015-05-18
Approximate non-ASCII characters

andrewgarner:
The font we use to render output documents has a limited character set so we need to remove unsupported characters.

### DIFF
--- a/app/models/transliterated_appointment_summary.rb
+++ b/app/models/transliterated_appointment_summary.rb
@@ -1,0 +1,9 @@
+class TransliteratedAppointmentSummary < SimpleDelegator
+  %i(first_name last_name guider_name
+     address_line_1 address_line_2 address_line_3
+     town county postcode country).each do |method_name|
+    define_method(method_name) do
+      I18n.transliterate(super())
+    end
+  end
+end
--- a/features/step_definitions/covering_letter_steps.rb
+++ b/features/step_definitions/covering_letter_steps.rb
@@ -3,17 +3,14 @@ Given(/^a customer has had a Pension Wise appointment$/) do
 end
 
 Then(/^it should include a covering letter$/) do
-  output_document = OutputDocument.new(@appointment_summary)
-  expected = {
-    attendee_address_line_1: output_document.attendee_address_line_1,
-    attendee_address_line_2: output_document.attendee_address_line_2,
-    attendee_address_line_3: output_document.attendee_address_line_3,
-    attendee_town: output_document.attendee_town,
-    attendee_county: output_document.attendee_county,
-    attendee_postcode: output_document.attendee_postcode,
-    attendee_name: output_document.attendee_name,
-    lead: output_document.lead
-  }
+  expected = fixture(:populated_csv).slice(%i(attendee_address_line_1
+                                              attendee_address_line_2
+                                              attendee_address_line_3
+                                              attendee_town
+                                              attendee_county
+                                              attendee_postcode
+                                              attendee_name
+                                              lead))
 
   expect_uploaded_csv_to_include(expected)
 end
--- a/features/step_definitions/record_of_guidance_contents_steps.rb
+++ b/features/step_definitions/record_of_guidance_contents_steps.rb
@@ -120,12 +120,7 @@ Given(/^(?:I|we) have captured the customer's details in an appointment summary$
 end
 
 Then(/^the record of guidance should include their details$/) do
-  output_document = OutputDocument.new(@appointment_summary)
-
-  expected = {
-    attendee_name: output_document.attendee_name,
-    value_of_pension_pots: output_document.value_of_pension_pots
-  }
+  expected = fixture(:populated_csv).slice(%i(attendee_name value_of_pension_pots))
 
   expect_uploaded_csv_to_include(expected)
 end
@@ -135,16 +130,12 @@ Given(/^(?:I|we) have captured appointment details in an appointment summary$/) 
 end
 
 Then(/^the record of guidance should include the details of the appointment$/) do
-  output_document = OutputDocument.new(@appointment_summary)
-
-  expected = {
-    appointment_date: output_document.appointment_date,
-    guider_first_name: output_document.guider_first_name,
-    guider_organisation: output_document.guider_organisation
-  }
+  expected = fixture(:populated_csv).slice(%i(appointment_date guider_first_name guider_organisation))
 
   expect_uploaded_csv_to_include(expected)
 
   appointment_reference = read_uploaded_csv.first[:appointment_reference]
-  expect(appointment_reference).to match(/^\d+#{output_document.appointment_reference}/)
+  expected_reference = %r{^\d+/#{@appointment_summary.reference_number}$}
+
+  expect(appointment_reference).to match(expected_reference)
 end
--- a/spec/models/transliterated_appointment_summary_spec.rb
+++ b/spec/models/transliterated_appointment_summary_spec.rb
@@ -1,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe TransliteratedAppointmentSummary do
+  transliterated_fields = %i(first_name
+                             last_name
+                             address_line_1
+                             address_line_2
+                             address_line_3
+                             town
+                             county
+                             postcode
+                             country
+                             guider_name)
+
+  transliterated_fields.each do |field|
+    describe ".#{field}" do
+      subject do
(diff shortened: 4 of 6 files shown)


spec/domain/output_document_spec.rb
```diff
@@ -180,10 +207,17 @@
     subject { output_document.lead }
 
     it do
-      is_expected.to eq(
-        'You recently had a Pension Wise guidance appointment with James ' \
-        "from The Pensions Advisory Service on #{appointment_date}."
-      )
+      is_expected.to eq('You recently had a Pension Wise guidance appointment with James ' \
+                        "from The Pensions Advisory Service on #{appointment_date}.")
```
stevenwilkin: This re-formating should strictly be included in this PR

stevenwilkin: *shouldn't

andrewgarner: OK. I'll follow the existing formatting in the new example introduced.

stevenwilkin:
Looks good, as long as we're ok with adding the transliteration into the domain object and not creating a new one to add in this new functionality

gjvis:
Can we pass in the transliterator, and test this?

gjvis:
And as @stevenwilkin suggests, perhaps the transliteration should happen in another object that wraps an `OutputDocument`?

andrewgarner:
The transliterator comes from `I18n` - it's already well tested.

stevenwilkin:
@gjvis is referring to the object that *applies* the transliteration, not the object *implementing* the transliteration @andrewgarner

gjvis:
I was, but I somehow missed that there were already specs on this PR. Need more coffee... :coffee: :coffee: :coffee:

gjvis:
I'm worried that `OutputDocument` is becoming a bit of a god-object, this does seem like the simplest thing to do to get this functionality in though, so might be too early to extract transliteration into its own thing. Happy to go with the consensus.

andrewgarner:
I've broken the concern of transliteration into a dedicated object and updated the acceptance tests to include transliteration.

cc/ @stevenwilkin @gjvis

features/step_definitions/covering_letter_steps.rb
```diff
@@ -3,17 +3,5 @@
...
-    attendee_address_line_3: output_document.attendee_address_line_3,
-    attendee_town: output_document.attendee_town,
-    attendee_county: output_document.attendee_county,
-    attendee_postcode: output_document.attendee_postcode,
-    attendee_name: output_document.attendee_name,
-    lead: output_document.lead
-  }
-
-  expect_uploaded_csv_to_include(expected)
+  # behaviour is outside of the system, nothing to assert
```
gjvis: I misunderstood this when we were discussing in person – the original assertion is still valid, we need to make sure the CSV contains all details needed to render a covering letter.

features/step_definitions/record_of_guidance_contents_steps.rb
```diff
@@ -120,12 +120,7 @@
 end
 
 Then(/^the record of guidance should include their details$/) do
-  output_document = OutputDocument.new(@appointment_summary)
-
-  expected = {
-    attendee_name: output_document.attendee_name,
-    value_of_pension_pots: output_document.value_of_pension_pots
-  }
+  expected = fixture(:populated_csv)
```
gjvis: we should assert only the fields we care about, not the entire csv row, otherwise unrelated changes will fail this assertion

features/step_definitions/record_of_guidance_contents_steps.rb
```diff
@@ -135,16 +130,12 @@
...
 
 Then(/^the record of guidance should include the details of the appointment$/) do
-  output_document = OutputDocument.new(@appointment_summary)
-
-  expected = {
-    appointment_date: output_document.appointment_date,
-    guider_first_name: output_document.guider_first_name,
-    guider_organisation: output_document.guider_organisation
-  }
+  expected = fixture(:populated_csv)
```
gjvis: we should assert only the fields we care about, not the entire csv row, otherwise unrelated changes will fail this assertion

spec/models/transliterated_appointment_summary_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  describe '.guider_name' do
+    subject { transliterated_appointment_summary.guider_name }
+
+    let(:guider_name) { 'García' }
+
+    it 'approximates non-ASCII characters' do
+      is_expected.to eq('Garcia')
+    end
+  end
```
gjvis: Some sort of shared example might be useful?

```ruby
shared_examples 'transliterates' do |field|
  subject do
    summary = AppointmentSummary.new(field => 'àéïöü')
    summary.public_send(field)
  end

  it 'approximates non-ASCII characters' do
    is_expected.to eq('aeiou')
  end
end


```

gjvis:
Structure is :+1:, much cleaner. Added a few small comments about tests

spec/models/transliterated_appointment_summary_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  describe '.first_name' do
+    subject { transliterated_appointment_summary.first_name }
+
+    let(:first_name) { 'Noemí' }
+
+    it 'approximates non-ASCII characters' do
+      is_expected.to eq('Noemi')
+    end
```
stevenwilkin: Better to use `it { is_expected.to eq(..) }` for all these assertions?

gjvis: My 2 cents: the example doesn't really communicate the intention, so I quite like the clarity of the message (in this case)

jiannis:
_Minor point_: might be worth testing out letters such as the [German ß](http://en.wikipedia.org/wiki/ß) (option + s) which doesn't have a direct one letter ASCII equivalent but should be transliterated as “ss”.

spec/models/transliterated_appointment_summary_spec.rb
gjvis: (danger: that code was GitHub-coded :skull: :warning:)

spec/models/transliterated_appointment_summary_spec.rb
stevenwilkin: My complaint was seeing `approximates non-ASCII characters` repeated down the file

features/step_definitions/covering_letter_steps.rb
andrewgarner: Fixed with b3c1075

features/step_definitions/record_of_guidance_contents_steps.rb
andrewgarner: Fixed with 8864534

features/step_definitions/record_of_guidance_contents_steps.rb
andrewgarner: Fixed with dd35f07

app/models/transliterated_appointment_summary.rb
```diff
@@ -0,0 +1,9 @@
+class TransliteratedAppointmentSummary < SimpleDelegator
+  %i(first_name last_name guider_name
+     address_line_1 address_line_2 address_line_3
+     town county postcode country).each do |method_name|
+    define_method(method_name) do
+      I18n.transliterate(super())
```
gjvis: if we leave off the `()` on `super()` then these will keep functioning if any of these methods ever accept arguments in future

features/step_definitions/covering_letter_steps.rb
```diff
@@ -3,17 +3,13 @@
...
-    attendee_name: output_document.attendee_name,
-    lead: output_document.lead
-  }
+  expected = fixture(:populated_csv).slice(%i(attendee_address_line_1
+                                              attendee_address_line_2
+                                              attendee_address_line_3
+                                              attendee_town
+                                              attendee_county
+                                              attendee_postcode
+                                              attendee_name))
```
gjvis: :+1: nice use of slice

app/models/transliterated_appointment_summary.rb
andrewgarner: Unfortunately that doesn't work, you get the following error `implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly.`

gjvis: interesting, didn't know that

gjvis:
Squish and merge? :+1: 

(I know yesterday you mentioned an additional change you wanted to make, but I can't remember what it was)

PR #168, opened 2015-05-20
Make login form fields wider

stevenwilkin:
To accommodate less technically savvy guiders with long email addresses.

How do you like that BEM @aduggin ?


stevenwilkin:
Updated with @aduggin's suggestions. Will merge when passes on Travis

PR #169, opened 2015-06-08
Ignore county in postcode search response

gjvis:
It is misleading and causing confusion amongst guiders. It is also not used for mail routing, and Ideal-Postcodes recommends not using it.


gjvis:
This is actively causing problems for guiders, and has been discussed over other channels. Merging!

andrewgarner:
:+1:

PR #172, opened 2015-06-17
Allow blank titles

gjvis:
It allows a little more flexibility, but adds the risk that a blank title and first_name leads to a "Dear Smith" salutation on a covering letter.


gjvis:
We've decided against this for now.

PR #175, opened 2015-07-28
Use latest govuk_admin_template

stevenwilkin:
GOV.UK Google Analytics can now be excluded so we don't need our fork:

https://github.com/alphagov/govuk_admin_template/pull/80


fofr:
:+1:

PR #176, opened 2015-08-13
 Use shell script rather than node module to notify Geckoboard

stevenwilkin:


stevenwilkin:
Updated to use pensionbot gist

PR #178, opened 2015-12-01
v2.0

stevenwilkin:


app/domain/output_document.rb
```diff
@@ -5,7 +5,9 @@ class OutputDocument
 
   delegate :id, :income_in_retirement, :continue_working, :unsure,
```
andrewgarner: We only need to delegate attributes that are needed for the output document. That means we can skip the circumstances and possibly pot size and income fields?

features/preview_record_of_guidance.feature
```diff
@@ -3,50 +3,17 @@ Feature: Record of guidance preview
...
   Then the sections that the preview includes should be (in order):
-    | introduction             |
-    | pension pot              |
+    | covering letter          |
+    | getting started          |
     | options overview         |
     | detail about each option |
-    | other information        |
-
-Scenario Outline: Guidance is tailored based on applicable circumstances
```
andrewgarner: We should have an acceptance test that proves that the applicable circumstances are captured in the database for analytical purposes.

```gherkin
Given "<circumstance>" applies to the customer
When I create their record of guidance
Then information about "<circumstance>" should be recorded
```

features/preview_record_of_guidance.feature
```diff
@@ -3,50 +3,17 @@ Feature: Record of guidance preview
...
-  Given one or more of the predefined circumstances applies to the customer
-  When I preview their record of guidance
-  Then the sections that the preview includes should be (in order):
-    | introduction                          |
-    | pension pot                           |
-    | options overview                      |
-    | detail about applicable circumstances |
-    | other information                     |
-
 Scenario: Generic record of guidance
```
andrewgarner: We should referred to them as 'standard' records not 'generic' records.

app/domain/output_document.rb
stevenwilkin: Good catch, I'm sure I'd removed them before but then I went crazy with the rebasing. I've changed the delegated attributes to just the id and the supplementary topics and everything's green.

features/preview_record_of_guidance.feature
stevenwilkin: Makes sense!

features/preview_record_of_guidance.feature
stevenwilkin: I've updated the scenario title

stevenwilkin:
Updated to take feedback into consideration.

Have included 1c874034147c0ee644703a712d40ff431d4ff17c with an acceptance test for the individual circumstances which are no longer part of the summary document but which we still want to store in the db.

features/record_of_guidance.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Record of guidance
```
andrewgarner: :+1: 

Can we also include a scenario(s) to cover the additional information we're collecting but not including in the document. Off the top of my head I can think of:

- `income_in_retirement`
- `value_of_pension_pots`
- `upper_value_of_pension_pots`
- `value_of_pension_pots_is_approximate`

stevenwilkin: Good call, these can probably be covered in the one scenerio

stevenwilkin: Updated

andrewgarner:
:+1:

PR #179, opened 2015-12-01
Slice the hash correctly

stevenwilkin:
`Hash#slice` requires a list of parameters, not an array.


andrewgarner:
Any ideas why the implementation works either way?

PR #181, opened 2015-12-01
Fix styleguide

stevenwilkin:
Maybe better to just remove the styleguide from this app?


andrewgarner:
This is what I did for the appointment summaries repo. Again though we need to remove the sensitive data.

PR #184, opened 2016-02-25
Add appointment types and option to select 50-54

benbarnett:
However until the printing partner is ready to process these documents, we will still request 'standard' documents from them.

The column addition will rebase out when https://github.com/guidance-guarantee-programme/output/pull/183 is merged.


config/initializers/assets.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.config.assets.paths << Rails.root.join('vendor', 'assets', 'bower_components')
-Rails.application.config.assets.precompile += %w( govuk_admin_template/favicon.png )
+Rails.application.config.assets.precompile += %w( govuk_admin_template/favicon.png
+                                                  govuk_admin_template/favicon-development.png )
```
andrewgarner: Maybe worth cherry-picking this out into a separate PR?

benbarnett: Good idea

db/migrate/20160225114042_add_apointment_type_to_appointment_summaries.rb
```diff
@@ -0,0 +1,5 @@
+class AddApointmentTypeToAppointmentSummaries < ActiveRecord::Migration
```
andrewgarner: We need to migrate the database before the code that depends on it goes into production. To do that we need to isolate the migration and schema changes into a separate PR. We can then make those changes in production.

spec/services/create_batch_spec.rb
```diff
@@ -39,7 +39,6 @@ def create_appointment_summary
       end
 
       include_examples 'ignore unsupported appointment summaries', format_preference: 'braille'
-      include_examples 'ignore unsupported appointment summaries', format_preference: 'large_text'
```
andrewgarner: We shouldn't be including large text in the batch yet since the upstream provider doesn't yet have that capability?

db/migrate/20160225114042_add_apointment_type_to_appointment_summaries.rb
andrewgarner: Ahh - I see #183

spec/services/create_batch_spec.rb
benbarnett: Good spot

benbarnett:
Tests will (should) fail until #183 is merged.

andrewgarner:
#183 is merged so you can rebase this.

features/appointment_type.feature
```diff
@@ -0,0 +1,14 @@
+Feature: Appointment type
+  As Pension Wise
+  We want to offer tailored appointments to customers
+  So we can provide relevant information where possible
+
+  Scenario Outline: Appointment types
+    Given the customer is given the <type> appointment
+    When we send them their record of guidance
+    Then it should be a <variant> summary document
```
andrewgarner: Shall we stick with the term "record of guidance" and then make a future change to bring all the domain language inline? It's a bit confusing to use "record of guidance" and "summary document" interchangeably.

benbarnett: Yeah, seems sensible

benbarnett: Changed

andrewgarner:
:+1:

PR #186, opened 2016-02-25
[DO NOT MERGE] Document templates 4.0.0 and foundation for 50-54 appointment types

benbarnett:
This PR will not let guiders select appointment types, but will update the documents to the latest content and lay the groundwork for when all organisations are ready to roll out the new appointment type feature.


andrewgarner:
We should mark this as [DO NOT MERGE] until we know the templates used by the upstream provider are inline with the preview templates.

PR #188, opened 2016-03-02
50-54 Appointments (including standard, large text and international)

benbarnett:
Adds ability for guiders to select appointment type, and updates to version 4 of the templates.


benbarnett:
@andrewgarner have taken out https://github.com/guidance-guarantee-programme/output/commit/ea2d69b as we'll just send all document types once the printers are ready.

app/services/create_batch.rb
```diff
@@ -1,7 +1,7 @@
 class CreateBatch
   def call
     unbatched_appointment_summaries = AppointmentSummary.unbatched
-                                      .where(format_preference: :standard)
+                                      .where.not(format_preference: 'braille')
```
andrewgarner: This isn't related to 50-54 appointments.

PR #189, opened 2016-03-07
Conditionally include or exclude attendee_country

benbarnett:
There is no need to include 'United Kingdom' on address labels when sending the documents. We therefore only include the country if the customer has specified a non-UK address.

@andrewgarner do you think we need more test coverage for the CSV phase?

### DIFF
--- a/app/domain/output_document.rb
+++ b/app/domain/output_document.rb
@@ -9,7 +9,7 @@ class OutputDocument
            :format_preference, :appointment_type,
            to: :appointment_summary
 
-  delegate :address_line_1, :address_line_2, :address_line_3, :town, :county, :postcode, :country,
+  delegate :address_line_1, :address_line_2, :address_line_3, :town, :county, :postcode,
            to: :appointment_summary, prefix: :attendee
 
   def initialize(appointment_summary)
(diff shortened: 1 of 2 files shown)


app/domain/output_document.rb
```diff
@@ -20,9 +20,11 @@ def attendee_name
     "#{appointment_summary.title} #{appointment_summary.first_name} #{appointment_summary.last_name}".squish
   end
 
-  def attendee_address
-    attendee_country = Countries.uk?(self.attendee_country) ? nil : self.attendee_country
+  def attendee_country
+    Countries.uk?(@appointment_summary.country) ? nil : @appointment_summary.country
```
andrewgarner: There is an [accessor method](https://github.com/guidance-guarantee-programme/output/blob/no-country-if-uk/app/domain/output_document.rb#L4) for `appointment_summary`. We should use that instead of the instance variable directly.

```rb
def attendee_country
  Countries.uk?(appointment_summary.country) ? nil : appointment_summary.country
end
```

benbarnett: Cool, makes sense

andrewgarner:
:+1:

PR #191, opened 2016-04-14
Appointment Summary Dashboard

benlovell:
Adds a basic appointment summary dashboard for displaying existing data in
tabular form. By default we display entries spanning back a month from the
current day, in reverse chronological order. The chosen date range can be
modified.

Only users with the new `admin` flag set to true can access this particular
area.

I'll be adding CSV export separately.

I'll squash down the commits after review.


features/step_definitions/appointment_summary_steps.rb
```diff
@@ -0,0 +1,35 @@
...
+  expect(@page).to have_appointments(count: Kaminari.config.default_per_page)
+end
+
+Then(/^I see there are multiple pages$/) do
+  expect(@page).to have_pages
+end
+
+Then(/^the date range defaults to the last month$/) do
+  expect(@page.start_date.value).to be_present
+  expect(@page.end_date.value).to be_present
```
dwhenry: I feel this is misleading to anyone reading the feature as the step definition does not match the implementation.

Feel like changing the step name here is the best approach rather than the implementation.

app/controllers/admin/appointment_summaries_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Admin
+  class AppointmentSummariesController < ::ApplicationController
+    before_action :authenticate_user!
+
+    def index
+      redirect_to :root unless current_user.admin?
```
dwhenry: might be nicer to wrap the two bit of authentication together into an `authenticate_admin!` method

benlovell:
@dwhenry Addressed those issues you raised.

benbarnett:
You beauty :+1:

PR #194, opened 2016-04-18
Digital summary

benbarnett:
This should be merged in accordance with the timelines of the rollout plan.

### DIFF
--- a/features/step_definitions/digital_summary_steps.rb
+++ b/features/step_definitions/digital_summary_steps.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+Given(/^the customer (.*) a digital appointment summary$/) do |requested|
+  @appointment_summary = fixture(:populated_appointment_summary).tap do |as|
+    as.requested_digital = requested_as_boolean(requested)
+  end
+end
+
+Then(/^we should know that the customer (.*) a digital version$/) do |requested|
+  appointment_summary = AppointmentSummary.first
+  expect(appointment_summary.requested_digital).to eq(requested_as_boolean(requested))
+end
+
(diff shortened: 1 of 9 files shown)


features/step_definitions/digital_summary_steps.rb
```diff
@@ -0,0 +1,18 @@
+Given(/^the customer (.*) a digital appointment summary$/) do |requested|
```
benlovell: We spoke about this in person but to summarise: this feels like a unit test, rather than something the end user could observe.

features/step_definitions/digital_summary_steps.rb
```diff
@@ -0,0 +1,18 @@
...
+end
+
+Then(/^we should know that the customer (.*) a digital version$/) do |requested|
+  requested_digital = case requested
+                      when 'requested' then true
+                      when 'did not request' then false
+                      end
+
+  appointment_summary = AppointmentSummary.first
+  expect(appointment_summary.public_send(:requested_digital)).to eq(requested_digital)
```
benlovell: Sending the `requested_digital` by reflection isn't necessary here since it's just a regular attribute with public visibility.

features/support/pages/appointment_summary_page.rb
```diff
@@ -140,6 +143,10 @@ def fill_in_format_preference(appointment_summary)
     end
   end
 
+  def fill_in_digital_request(appointment_summary)
```
benlovell: You might consider this overkill given this is used once and requires more code than just inlining this method. Although, I appreciate there is an emergent pattern being followed here.

benlovell:
:shipit: 🚀

PR #196, opened 2016-04-18
Rake tasks for User administration

benlovell:
Adds two rake tasks. One for listing existing users with the administrative
flag set and another for toggling the flag on a user specified by the provided
email address.


benbarnett:
👍

PR #207, opened 2016-05-10
[WIP] Tag logs with devise's `current_user`

benlovell:
This adds `[uid: 1]` prefixes to logging via rails tagged logging. This
is so we can track a user's steps through the application flow to check
if they've been recorded post-interview summaries correctly.

This is done via a rack middleware since using the regular
`config.log_tags` with a `proc` to retrieve information from cookies /
session is too far down the middleware chain, after
`Rails::Rack::Logger` is called. /shrug


mattlucht:
:+1 by @dwhenry

PR #208, opened 2016-05-11
Abstract CSV generation

benlovell:
Uses the recently introduced `CsvGenerator` from our reporting project.
This is good for now while I extract a gem to share between both
projects.


dwhenry:
🚢

PR #210, opened 2016-07-13
Add number_of_appointments field to DB

dwhenry:
-- DO NOT MERGE --
until the related code PR #211 has been approved


db/migrate/20160712095513_add_number_of_appointments_to_appointment_summary.rb
```diff
@@ -0,0 +1,5 @@
+class AddNumberOfAppointmentsToAppointmentSummary < ActiveRecord::Migration
+  def change
+    add_column :appointment_summaries, :number_of_previous_appointments, :string, null: false, default: ''
```
benlovell: Any specific reason this is a string and not numeric?

PR #213, opened 2016-07-15
Appendix c fields

dwhenry:
-- DO NOT MERGE --

- [ ] Merge and migrate DB changes on production
- [ ] Update PR to merge into master

Add count of pension pots field and add missing appendix c field to exported CSV


db/migrate/20160715124444_add_count_of_pension_pots.rb
```diff
@@ -0,0 +1,5 @@
+class AddCountOfPensionPots < ActiveRecord::Migration
+  def change
+    add_column :appointment_summaries, :count_of_pension_pots, :integer
```
benlovell: No default? Nullable?

dwhenry: Yep, this was intentional as any default could give false data, and null allows then to skip the field if they don't know the answer.

benlovell:
👍

PR #216, opened 2016-07-25
[DO NOT MERGE] Email notification of digital version

dwhenry:
-- DO NOT MERGE -- 

Ensure DB migrations are merged and run against production first

Request optional email field during summary document generation. If the email is provided and a digital version was requested ask `Notify` to send an email to the user.


benbarnett:
As we're doing a few of these 'don't deploy yet' PR's, it might be worth putting `[DO NOT MERGE]` in the name of the PR itself just in case.

app/models/appointment_summary.rb
```diff
@@ -4,6 +4,10 @@
 
 class AppointmentSummary < ActiveRecord::Base
   DIGITAL_BY_DEFAULT_START_DATE = Date.new(2016, 6, 21)
+  EMAIL_REGEXP = Regexp.union(
+    /\A[^\s][^@]+@([^@\.]+\.)+[^@\.]+[^\s]\z/,
+    /\A\z/
+  )
```
benbarnett: Don't know how far we want to take this stuff but this allows in special characters too:

```ruby
[11] pry(main)> "t^^^@hiy$.com".match(EMAIL_REGEXP)
=> #<MatchData "t^^^@hiy$.com" 1:"hiy$.">
[13] pry(main)> "t£$%^&*()@test.com".match(EMAIL_REGEXP)
=> #<MatchData "t£$%^&*()@test.com" 1:"test.">
```

benbarnett: As discussed let's just keep this loose for now and keep an eye on the dashboard for failed-to-deliver emails.

features/digital_summary.feature
```diff
@@ -12,3 +12,13 @@ Feature: Digital Appointment Summaries
       | requested_digital |
       | requested         |
       | did not request   |
+
+  Scenario: Email notification of summary document location
+    Given the customer requested a digital appointment summary
+    When I create their record of guidance
```
benbarnett: We need to go through this project at some point and change 'record of guidance' to 'appointment summary' – they both mean the same thing.

features/digital_summary.feature
```diff
@@ -12,3 +12,13 @@ Feature: Digital Appointment Summaries
...
+    Then the customer should be notified
```
benbarnett: "Then the customer should be notified __by email__"?

features/digital_summary.feature
```diff
@@ -12,3 +12,13 @@ Feature: Digital Appointment Summaries
...
+
+  Scenario: No email notification when postal notification requested
+    Given the customer did not requested a digital appointment summary
```
benbarnett: Maybe this should be `Given the customer requested an appointment summary by post`?

It's about selecting that method rather than not selecting digital... if that makes sense.

features/digital_summary.feature
```diff
@@ -12,3 +12,13 @@ Feature: Digital Appointment Summaries
...
+    When I create their record of guidance
+    Then the customer should not be notified
```
benbarnett: Feels like this is getting into unit testing territory. Should this be more like:

"Then the appointment summary is sent by post"

and then we ensure it's landing in the CSV

benbarnett: Having just spoken in person about this it seems the best approach is to rename this entire feature to 'Delivery methods' which then positively tests the behaviour of each delivery method (digital/postal).

For testing the inverse of these (i.e. what happens if NOT digital), then this should probably be done at a unit test level.

benlovell:
:shipit: Nice.

PR #219, opened 2016-08-10
Search appointment by surname and reference

dwhenry:
Add additional search options to admin interface.
Search does a case insensitive wildcard like match
using the value provided.

### DIFF
--- a/app/forms/appointment_summary_browser.rb
+++ b/app/forms/appointment_summary_browser.rb
@@ -39,7 +39,7 @@ class AppointmentSummaryBrowser
     return scope unless @search_string.present?
 
     scope.where(
-      %(appointment_summaries.last_name ilike ? OR appointment_summaries.reference_number = ?),
+      'appointment_summaries.last_name ilike ? OR appointment_summaries.reference_number = ?',
       "%#{@search_string}%",
       @search_string
     )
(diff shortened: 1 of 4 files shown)


app/forms/appointment_summary_browser.rb
```diff
@@ -26,4 +30,18 @@ def paginated_results
...
+
+  def date_filter(scope)
+    scope.where(date_of_appointment: start_date..end_date)
+  end
+
+  def text_filter(scope)
+    return scope unless @search_string.present?
+
+    scope.where(
+      %(appointment_summaries.last_name ilike ? OR appointment_summaries.reference_number = ?),
```
benlovell: Any reason you chose not to simply quote this string?

spec/forms/appointment_summary_browser_spec.rb
```diff
@@ -33,7 +70,8 @@
           described_class.new(
             page: nil,
             start_date: '15/01/2016',
-            end_date: '16/01/2016'
+            end_date: '16/01/2016',
+            search_string: nil
```
benlovell: Maybe default the kwarg to `nil` so we don't need this here and elsewhere?

dwhenry: A value is only passed in here, everywhere else a value is used.

PR #223, opened 2016-08-22
Send Ineligible template if customer is ineligible

dwhenry:
Previously guiders could only send notifications 
to ineligible customers via post

This release will correspond with an update to guider notes, allowing them to send ineligible notifications via email if the customer opts into digital.

### DIFF
--- a/spec/jobs/notify_via_email_spec.rb
+++ b/spec/jobs/notify_via_email_spec.rb
@@ -8,7 +8,8 @@ RSpec.describe NotifyViaEmail do
       :config,
       service_id: 'service',
       secret_id: 'secret',
-      appointment_summary_template_id: 'template'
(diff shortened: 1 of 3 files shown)


spec/jobs/notify_via_email_spec.rb
```diff
@@ -18,22 +19,46 @@
...
+          }
+        }.to_json
+      )
 
-    described_class.perform_now(appointment_summary, config: config)
+      described_class.perform_now(appointment_summary, config: config)
+    end
+  end
+
+  context 'when the customer does not have DC pension' do
```
benlovell: "a DC pension" would read slightly better here.

dwhenry: Good point. also updated the it statements to be clearer of purpose.

PR #224, opened 2016-09-02
Regenerate summary document

dwhenry:
Allow Team leaders to regenerate a summary document. This creates a new summary document record with allow any of the details to be corrected if required.

The PR also removes the admin flag and replaces it with a role, at this stage users can only have a single role at any one time.

### DIFF
--- a/app/forms/appointment_summary_finder.rb
+++ b/app/forms/appointment_summary_finder.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+class AppointmentSummaryFinder
+  include ActiveModel::Model
+
+  attr_accessor :search_string
+
+  def initialize(search_string:)
+    @search_string = search_string
+  end
+
+  def results
+    return [] if search_string.blank?
+
+    AppointmentSummary
+      .where(reference_number: search_string)
+      .order(:date_of_appointment, :created_at)
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,11 +24,11 @@ class User < ActiveRecord::Base
   end
 
   def admin?
-    role == ADMIN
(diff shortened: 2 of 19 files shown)


app/forms/appointment_summary_finder.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+class AppointmentSummaryFinder
+  include ActiveModel::Model
+
+  attr_accessor :search_string
```
benlovell: You're not using this it seems?

app/models/user.rb
```diff
@@ -1,11 +1,19 @@
 # frozen_string_literal: true
 class User < ActiveRecord::Base
+  deprecated_columns :admin
```
benlovell: Nice 👍

benlovell:
I think this PR is telling us we need this project under signon 😄

dwhenry:
@benlovell no argument from me on that. 

I think it is more of timing and resources as: 

* I believe the plan is to kill this once we have only booking/scheduling in place.
* Moving login across will take time and coordination with TPAS

app/forms/appointment_summary_finder.rb
dwhenry: removed

dwhenry: Added back it as it is used in the view.. :(

benlovell: Replace the ivar accesses with this instead then perhaps?

dwhenry: Went the other direction as used the accessor for everything. :)

benlovell: That's what I meant. 👍

PR #230, opened 2016-10-05
Allow resend with the same email address

dwhenry:
* Don't raise NotificationAlreadySent error when trying to resend notification without changing the email address
* Cleaner message for NotificationAlreadySent - only return useful variables 
* Update email regex - use the regex used by notify

### DIFF
--- a/app/controllers/admin/appointment_summaries_controller.rb
+++ b/app/controllers/admin/appointment_summaries_controller.rb
@@ -47,6 +47,7 @@ module Admin
     def appointment_summary_params
       params.require(:appointment_summary)
             .permit(:email)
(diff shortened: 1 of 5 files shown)


app/controllers/admin/appointment_summaries_controller.rb
```diff
@@ -47,6 +47,7 @@ def authenticate_admin!
     def appointment_summary_params
       params.require(:appointment_summary)
             .permit(:email)
+            .merge('notification_id' => nil)
```
benlovell: Any reason this isn't 1.9 style?

dwhenry:
@benlovell you happy with this one now?

PR #233, opened 2016-10-12
Update email regex

dwhenry:
Keep it inline with the Notify regex to avoid predictable
errors when sending the email across.

Difference to the notify regex:

* block '..' in regex - notify has a secondary check
* block spaces in email address

### DIFF
--- a/app/models/appointment_summary.rb
+++ b/app/models/appointment_summary.rb
@@ -4,8 +4,9 @@ require 'uri'
 
 class AppointmentSummary < ActiveRecord::Base
   DIGITAL_BY_DEFAULT_START_DATE = Date.new(2016, 6, 21)
+  # bassed off: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/recipients.py#L22
   EMAIL_REGEXP = Regexp.union(
-    /\A[^\@\s]+@[^\@\.\s]+(\.[^\@\.\s]+)+$/,
+    /\A([^";@\s\.]+\.)*[^";@\s\.]+@([^";@\s\.]+\.)+[a-z]{2,10}\z/,
     /\A\z/
   )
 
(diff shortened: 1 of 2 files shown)


app/models/appointment_summary.rb
```diff
@@ -4,8 +4,9 @@
 
 class AppointmentSummary < ActiveRecord::Base
   DIGITAL_BY_DEFAULT_START_DATE = Date.new(2016, 6, 21)
+  # bassed off: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/recipients.py#L22
```
benbarnett: 🎺

benbarnett: 🐟

AndrewVos: < ActiveRecord::Bass

app/models/appointment_summary.rb
```diff
@@ -4,8 +4,9 @@
...
   EMAIL_REGEXP = Regexp.union(
-    /\A[^\@\s]+@[^\@\.\s]+(\.[^\@\.\s]+)+$/,
+    /\A([^";@\s\.]+\.)*[^";@\s\.]+@([^";@\s\.]+\.)+[a-z]{2,10}\z/,
```
AndrewVos: Couldn't we just switch to using an email field here:

https://github.com/guidance-guarantee-programme/output/blob/master/app/views/appointment_summaries/new.html.erb#L39

I'm not really a fan of whitelisting email formats as we go along, we'll potentially lose customers?

AndrewVos: Browser support is pretty good: http://caniuse.com/#feat=input-email-tel-url

dwhenry: I think in this case it is fine as we are not storing the email address to create a customer, but to immediately send a notification email (optional). 

The regex is taken from the system that sends the notification so it would break anyway - just later in a background job and then require it to be manually reviewed.

Added to which the email is being entered by a third party - so scope for error.

The use case here is to flag known issues in the email address before the hit the notify system.

AndrewVos: Fair enough.

PR #235, opened 2016-10-14
[DO NOT MERGE] Replace Devise with GDS::SSO

dwhenry:
TPAS users now have a SSO login so we should be
using this to authenticate


app/views/layouts/application.html.erb
```diff
@@ -21,18 +21,17 @@
...
-      <span class='session-details-panel__username'>Logged in as: <%= current_user.name %></span>
-      <%= link_to 'Logout', destroy_user_session_path %>
-      <%= link_to 'Change password', edit_user_registration_path %>
+      <%= link_to current_user.name, Plek.current.find('signon') %>
+      &bull; <%= link_to 'Sign out', gds_sign_out_path %>
     </div>
   <% end %>
 
-  <% if current_user.team_leader? %>
+  <% if current_user.has_permission?('team_leader') %>
```
benlovell: These might as well be the same predicates as before. Seems a little cleaner to keep the role name hidden away as knowledge of the user.

app/views/layouts/application.html.erb
```diff
@@ -21,18 +21,17 @@
...
     <% content_for :navbar_items do %>
       <li>
         <%= link_to 'Reprint', appointment_summaries_path %>
       </li>
-      <% if current_user.admin? %>
+      <% if current_user.has_permission?('analyst') %>
```
benlovell: Same.

app/views/layouts/application.html.erb
```diff
@@ -21,18 +21,17 @@
 <% if user_signed_in? %>
   <%= content_for :navbar_right do %>
     <div class='session-details-panel'>
-      <span class='session-details-panel__username'>Logged in as: <%= current_user.name %></span>
-      <%= link_to 'Logout', destroy_user_session_path %>
-      <%= link_to 'Change password', edit_user_registration_path %>
+      <%= link_to current_user.name, Plek.current.find('signon') %>
```
benlovell: `govuk_admin_template` has a config setting that when enabled does this automatically for us.

dwhenry: This requires the GovukAdminTemplate to be updated and breaks the error pages. As a result I would prefer to do this as a separate PR once this one is mreged. :)

app/views/layouts/application.html.erb
dwhenry: done and done

app/views/layouts/application.html.erb
benlovell: Cool, makes sense.

benlovell:
We should be sure to give this a thorough going over in staging. 👍

dwhenry:
@benlovell can you please have a look over this again to confirm you are happy.

Aman added an extra request to remove the guider_organisation question as it adds no value and has been shoe horned into the PR.

PR #237, opened 2016-10-20
[DO NOT MERGE] Add retirement income fields

dwhenry:
List of options for the type of income the customer
is expecting in retirement.

We have also remove the question asking if you
have addiitonal income in retirement as it is cover
by the results of this question.

### DIFF
--- a/spec/services/appointment_summary_csv_spec.rb
+++ b/spec/services/appointment_summary_csv_spec.rb
@@ -35,7 +35,6 @@ RSpec.describe AppointmentSummaryCsv do
           supplementary_debt
           supplementary_ill_health
           supplementary_defined_benefit_pensions
-          income_in_retirement
           continue_working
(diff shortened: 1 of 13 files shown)


spec/services/appointment_summary_csv_spec.rb
```diff
@@ -49,6 +48,15 @@
           created_at
           email
           notification_id
+          retirement_income_other_state_benefits
```
benbarnett: Do we still need this whole spec?

dwhenry: I have no strong feelings either way. It is not adding a great deal of value, but at the same time it does ensure the CSV output is correct.

benbarnett:
Sorry was getting confused, I thought this was the spec for the Williams
Lea stuff

PR #243, opened 2016-12-12
Remove William's Lee related code

dwhenry:
## manual testing completed

This is old code used to split appointments into
batches and send each batch of appointments to
Williams Lee via SFTP.


features/accessibility.feature
```diff
@@ -5,18 +5,11 @@ Feature: Accessibility
 
 Scenario Outline: Document formats
   Given the customer prefers to receive documentation in <format> format
-  When we send them documentation
+  When we send them their record of guidance
```
benlovell: Can we drop the 'them' here as it's superfluous.

features/step_definitions/covering_letter_steps.rb
```diff
@@ -4,14 +4,5 @@
...
-                                               attendee_address_line_2
-                                               attendee_address_line_3
-                                               attendee_town
-                                               attendee_county
-                                               attendee_postcode
-                                               attendee_name
-                                               lead))
-
-  expect_uploaded_csv_to_include(expected)
+  expect(AppointmentSummary.last).to be_present
```
benlovell: Is this specific enough given the specificity of the check it replaces?

features/step_definitions/record_of_guidance_contents_steps.rb
```diff
@@ -10,18 +10,15 @@
...
 
 Given(/^(?:I|we) have captured the customer's details in an appointment summary$/) do
   @appointment_summary = fixture(:populated_appointment_summary)
 end
 
 Then(/^the record of guidance should include their details$/) do
-  expected = fixture(:populated_csv).slice(:attendee_name)
-
-  expect_uploaded_csv_to_include(expected)
+  puts @appointment_summary.slice(:first_name, :last_name)
```
benlovell: Kill this please.

features/accessibility.feature
dwhenry: If we want to do this it should be a different PR as it is not related to removing Williams Lee and would affect multiple features.

The change I have made which is a simple rationalisation of the steps as `we send them documentation` was just calling `we send them their record of guidance`

benlovell: Gotcha. Leave as-is for now then.

dwhenry:
@benlovell if you approve this I'll deploy. :)

PR #244, opened 2016-12-20
Rails 5 upgrade

dwhenry:


Gemfile
```diff
@@ -24,9 +24,9 @@ gem 'notifications-ruby-client', github: 'alphagov/notifications-ruby-client'
 gem 'output-templates', '~> 4.4.4', github: 'guidance-guarantee-programme/output-templates'
 gem 'pg'
 gem 'plek'
-gem 'princely'
+gem 'princely', github: 'guidance-guarantee-programme/princely', branch: 'remove-alias-method-chain'
 gem 'puma'
-gem 'rails', '4.2.6'
+gem 'rails', '5.0.0.1'
```
benlovell: Can we also bump this to 5.0.1 now?

config/environments/development.rb
```diff
@@ -10,13 +9,28 @@
...
+  # Show full error reports.
   config.consider_all_requests_local = true
-  config.action_controller.perform_caching = false
+
+  # Enable/disable caching. By default caching is disabled.
+  if Rails.root.join('tmp/caching-dev.txt').exist?
+    config.action_controller.perform_caching = true
+
+    config.cache_store = :memory_store
+    config.public_file_server.headers = {
```
benlovell: We ought to add this to the production environment too.

db/migrate/20161207095801_remove_permissions_default_value.rb
```diff
@@ -0,0 +1,5 @@
+class RemovePermissionsDefaultValue < ActiveRecord::Migration[5.0]
+  def change
+    change_column :users, :permissions, :string, default: nil
```
benlovell: Why did we need to do this?

dwhenry: It's to do with the YAML deserialising. 
Without the field defaults to an empty string which does NOT deserialise to an Array as expected.

config/environments/development.rb
dwhenry: `public_file_server` is not being used in prod as `RAILS_SERVE_STATIC_FILES` is not set.

Gemfile
dwhenry: doing..

config/environments/development.rb
benlovell: How are we serving the assets?

db/migrate/20161207095801_remove_permissions_default_value.rb
benlovell: Did this manifest itself in the specs?

dwhenry: yes

benlovell: Cool.

dwhenry:
@benlovell all done including use of ApplicationRecord.

dwhenry:
@benlovell all changes made as requested.

PR #246, opened 2017-01-20
Create summary document activity in TAP

dwhenry:
Once a summary document has been created trigger
the appropriate activity to be created in TAP.

the following variables will need to be set:

```ruby
TELEPHONE_APPOINTMENTS_API_BEARER_TOKEN
TELEPHONE_APPOINTMENTS_API_URI
```

**Pending** Fix build pack

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem 'retriable'
 gem 'sass-rails', '~> 5.0'
 gem 'sidekiq'
 gem 'sinatra', require: nil # Sidekiq UI
+gem 'telephone_appointments'
 gem 'uglifier', '>= 1.3.0'
 gem 'uk_postcode'
 
--- a/app/jobs/create_tap_activity.rb
+++ b/app/jobs/create_tap_activity.rb
@@ -1,0 +1,23 @@
+require 'notifications/client'
+
+class CreateTapActivity < ApplicationJob
+  class UnableToCreateSummaryDocumentActivity < StandardError; end
+
+  rescue_from(CreateTapActivity::UnableToCreateSummaryDocumentActivity) do |exception|
+    Bugsnag.notify(exception)
+  end
+
+  queue_as :default
+
+  def perform(appointment_summary, owner)
+    telephone_appointment = TelephoneAppointments::SummaryDocumentActivity.new(
+      appointment_id: appointment_summary.reference_number,
+      owner_uid: owner.uid,
+      delivery_method: appointment_summary.requested_digital ? 'digital' : 'postal'
+    )
+
+    return if telephone_appointment.save
+
+    raise UnableToCreateSummaryDocumentActivity, "Errors: #{telephone_appointment.errors.inspect}"
+  end
+end
--- a/spec/jobs/create_tap_activity_spec.rb
+++ b/spec/jobs/create_tap_activity_spec.rb
@@ -1,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe CreateTapActivity do
+  let(:owner) { double(:owner, uid: 'aaaaaa') }
+  before do
+    allow(TelephoneAppointments::SummaryDocumentActivity).to receive(:new).and_return(summary_document_activity)
+  end
+
+  context 'when a posted summary document is requested' do
+    let(:appointment_summary) { double(:appointment_summary, reference_number: '1234', requested_digital: false) }
+    let(:summary_document_activity) { double(:summary_document_activity, save: true) }
+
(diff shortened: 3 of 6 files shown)


Gemfile
```diff
@@ -32,6 +32,7 @@ gem 'retriable'
 gem 'sass-rails', '~> 5.0'
 gem 'sidekiq'
 gem 'sinatra', require: nil # Sidekiq UI
+gem 'telephone_appointments', github: 'guidance-guarantee-programme/telephone_appointments'
```
benlovell: What's stopping us from publishing this gem?

dwhenry: Updated to source from rubygems instead of github

features/step_definitions/delivery_methods_steps.rb
```diff
@@ -36,7 +36,7 @@
 end
 
 Then(/^the customer should be notified by email$/) do
-  last_job = ActiveJob::Base.queue_adapter.enqueued_jobs.last
+  last_job = ActiveJob::Base.queue_adapter.enqueued_jobs.reverse.detect { |job| job[:job] == NotifyViaEmail }
```
benlovell: This feels like absolutely the wrong thing to do with cucumber. Is there a precedent for this style of testing here already?

dwhenry: I can change the order the jobs are created in and then still use last. but that feels fragile. That is why I went with the approach of getting the most recent job of the required type.

benlovell: It wasn't about the approach for retrieving the job, it was more the fact you're digging way down into jobs from the perspective of a user-facing test. I'd favour checking on something in the presence of the page the *user* sees as confirmation.

You could then check a job is enqueued from the perspective of a unit test around this behaviour.

app/jobs/create_tap_activity.rb
```diff
@@ -0,0 +1,20 @@
...
+    telephone_appointment = TelephoneAppointments::SummaryDocumentActivity.new(
+      appointment_id: appointment_summary.reference_number,
+      owner_uid: guider.uid,
+      delivery_method: appointment_summary.requested_digital ? 'digital' : 'postal'
+    )
+
+    return if telephone_appointment.save
+
+    errors = telephone_appointment.errors.map { |field, error| "#{field}: #{error}" }.join(', ')
+    raise UnableToCreateSummaryDocumentActivity, "Errors: #{errors}"
```
benlovell: What does `errors.inspect` output? That may be sufficient.

app/jobs/create_tap_activity.rb
```diff
@@ -0,0 +1,20 @@
...
+class CreateTapActivity < ApplicationJob
+  class UnableToCreateSummaryDocumentActivity < StandardError; end
+
+  queue_as :default
+
+  def perform(appointment_summary, guider)
+    telephone_appointment = TelephoneAppointments::SummaryDocumentActivity.new(
+      appointment_id: appointment_summary.reference_number,
+      owner_uid: guider.uid,
+      delivery_method: appointment_summary.requested_digital ? 'digital' : 'postal'
```
benlovell: I can't help feeling this API should have also accepted `requested_digital: true|false`... and done any conversion internally.

app/jobs/create_tap_activity.rb
benlovell: Also, this will just keep retrying - which doesn't make sense if we get validation errors back - as it won't be recoverable without intervention. The right thing to do here is report the error to bugsnag so we can follow up there.

app/jobs/create_tap_activity.rb
```diff
@@ -0,0 +1,23 @@
+require 'notifications/client'
+
+class CreateTapActivity < ApplicationJob
+  class UnableToCreateSummaryDocumentActivity < StandardError; end
+
+  rescue_from(ActiveRecord::RecordNotFound) do |exception|
```
benlovell: This is meant to be your custom exception right?

spec/jobs/create_tap_activity_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      described_class.perform_now(appointment_summary, owner)
+    end
+  end
+
+  context 'when the summary document fails to save' do
+    let(:summary_document_activity) do
+      double(:summary_document_activity, save: false, errors: { 'owner_id' => 'can be black' })
+    end
+
+    it 'creates a new summary document activity' do
```
benlovell: This isn't true.

PR #254, opened 2017-02-16
Ensure appointments are summarised via TAP

benlovell:
We should stop guiders from landing directly here and summarising
appointments by entering the TAP reference. When this happens, they
circumvent the status checks that occur in TAP and thus the outbound
webhook to create the TAP activity fails upon creation.

This change throws up a warning screen to the guider (team leaders are
still permitted) telling them to summarise the appointment from TAP.

screen shot 2017-02-16 at 15 46 53

### DIFF
--- a/app/controllers/appointment_summaries_controller.rb
+++ b/app/controllers/appointment_summaries_controller.rb
@@ -11,12 +11,16 @@ class AppointmentSummariesController < ApplicationController
   end
 
   def new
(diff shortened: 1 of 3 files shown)


Guntrisoft:
👍

alexkouz:
@benlovell Suggested text:
Create appointment summary in planner
After the appointment status is updated use the Telephone Appointment Planner to create the summary.

PR #256, opened 2017-02-21
Depend on gem release for notify client

benlovell:
We were previously depending on the outdated git-based gemspec.

The update introduced a breaking change around the client arguments.
The service ID is now inferred from the secure token so this required
changes to suit.


Guntrisoft:
👍

PR #258, opened 2017-02-22
Poll for notify delivery statuses

benlovell:
We need to poll for notify email delivery statuses to ensure we remain
deliverable at all times. Unfortunately notify does not supply webhooks
so we have to poll for the statuses instead.

A summary will be included for polling when it's less than two days old,
has a notify supplied ID and its `notify_completed_at` timestamp is
absent. When notify deems itself 'done' with our notification whether
due to a permanent failure or delivery it will supply us with the
`completed_at` timestamp that we use to mark the summary for exclusion
next time around.

This will need to be scheduled `rails runner 'NotifyDeliveryChecker.new.call'`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,3 +32,8 @@ Style/RescueModifier:
 
 Style/EmptyMethod:
   Enabled: false
+
+Metrics/AbcSize:
+  Exclude:
+    - spec/**/*
+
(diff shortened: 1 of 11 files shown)


benlovell:
[As a guider I can see the email delivery status for a digital summary document.](https://trello.com/c/wsdaQLBv/14-as-a-guider-i-can-see-the-email-delivery-status-for-a-digital-summary-document)

benlovell:
[As a guider I can see the email delivery status for a digital summary document.](https://trello.com/c/wsdaQLBv/14-as-a-guider-i-can-see-the-email-delivery-status-for-a-digital-summary-document)

Guntrisoft:
👍

PR #260, opened 2017-02-23
Ensure we pass the correct reference to the API

benlovell:
We need to pass the original booking reference to the API so we can
reconcile correctly between the two records.


Guntrisoft:
👍

benlovell:
[As a guider I can see a email delivery failure for a digital summary document](https://trello.com/c/wsdaQLBv/14-as-a-guider-i-can-see-a-email-delivery-failure-for-a-digital-summary-document)

PR #262, opened 2017-02-24
Check for notify deliveries every minute

benlovell:
This is a bit of a hack to workaround the heroku scheduler's limit on
running jobs at most once every ten minutes. When this newly introduced
rake task runs it will first schedule a check to run immediately, then
another 9 checks, covering each subsequent minute.


benlovell:
[As a guider I can see a email delivery failure for a digital summary document](https://trello.com/c/wsdaQLBv/14-as-a-guider-i-can-see-a-email-delivery-failure-for-a-digital-summary-document)

PR #267, opened 2017-03-07
Permit regeneration of TAP appointment summaries

benlovell:
Lets guiders modify existing appointment summaries when the original
appointment record is updated in TAP. The previously persisted summary
details are bound to the form and are modifiable. When validated /
persisted we ping TAP to create another activity entry and if necessary
email the customer.


benlovell:

Guntrisoft:
👍

PR #268, opened 2017-03-07
Don't guard against resending of digital summary

benlovell:
Now we can reissue the summaries we don't need to guard against
resending them.


benlovell:

Guntrisoft:
👍

PR #269, opened 2017-03-08
Ensure we poll for delivery status on resends

benlovell:
We need to reset the timestamp and status to ensure once we have
reissued an appointment summary digitally via notify that we then poll
at the scheduled intervals for its delivery status. Prior to this
change, we were not resetting those markers against the record so it was
being skipped when determining which records to poll.


benlovell:

PR #270, opened 2017-03-08
Remove email confirmation step in creating summary document

Guntrisoft:
Now that the email address is read only (coming from TAP)
there is no need for the separate email confirmation
step in the process, so this removes it.


benlovell:
Nicely.

Guntrisoft:
[As a guider I am not able to edit pre populated fields in the summary document generator](https://trello.com/c/NPu5XEtD/27-as-a-guider-i-am-not-able-to-edit-pre-populated-fields-in-the-summary-document-generator)

PR #272, opened 2017-03-10
Ensure previews are displayed when regenerating

benlovell:
Resolves a bug whereby previews were being skipped for postal summaries
when regenerating via TAP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
 cache:
   bundler: true
   directories:
+  - '.downloads'
   - node_modules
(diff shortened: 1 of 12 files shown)


benlovell:
[As a guider when I regenerate a postal version of the summary document  I want to see a PDF version so that I am able to print it](https://trello.com/c/nfM3dWVW/49-as-a-guider-when-i-regenerate-a-postal-version-of-the-summary-document-i-want-to-see-a-pdf-version-so-that-i-am-able-to-print-it)

PR #279, opened 2017-03-20
Add supplementary information to the confirmation screen

Guntrisoft:
This will display what supplementary info was selected
on the previous screen when creating the summary document.

This helps the guiders understand what will be displayed
in the summary document when sent out/printed

screen shot 2017-03-20 at 12 10 40


Guntrisoft:
[As a guider I want to confirm which supplementary information has been selected for the customer before I generate the summary document](https://trello.com/c/irKZJmtt/67-as-a-guider-i-want-to-confirm-which-supplementary-information-has-been-selected-for-the-customer-before-i-generate-the-summary-d)

PR #280, opened 2017-04-01
Specify appointment type via TAP

benlovell:
Makes these radio options disabled since their values will now be
provided by TAP. I removed the feature as it's now redundant given the
options are no longer modifiable.


benlovell:
[As a guider I want the option to select if the appointment is a 50-54 or 55+ type of appointment](https://trello.com/c/hsR0bRze/90-as-a-guider-i-want-the-option-to-select-if-the-appointment-is-a-50-54-or-55-type-of-appointment)

benlovell:
Deployment should be coordinated with guidance-guarantee-programme/telephone_appointment_planner#502

Guntrisoft:
👍

PR #286, opened 2017-06-14
Loosen validation on reference number to allow any string

jongilbraith:
Appointments booked via other systems have alphanumeric reference numbers, so this allows sharing of the same values to be used for summary documents.


jongilbraith:
[As a f2f guider generating a summary document I am able to input an alphanumeric booking reference number](https://trello.com/c/stKEBmrq/200-as-a-f2f-guider-generating-a-summary-document-i-am-able-to-input-an-alphanumeric-booking-reference-number)

PR #288, opened 2017-06-15
Add Mailgun email validation to summary form

jongilbraith:
Much the same as we've seen before in the other apps.

Validation capture

Styles recycled from `pension_guidance` with minor tweaking, but lead to replacing `sass-rails` with `sassc-rails` (as used in `pension_guidance`) because of unsupported syntax (definition if mixins within control flow statements).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,64 +1,68 @@
-# frozen_string_literal: true
-source 'https://rubygems.org'
-
 ruby IO.read('.ruby-version').strip
 
 # force Bundler to use HTTPS for github repos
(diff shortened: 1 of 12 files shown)


jongilbraith:
[As a f2f guider when generating a summary document I am given suggestions of valid email address formats when entering invalid emails addresses](https://trello.com/c/4hVlQkgC/199-as-a-f2f-guider-when-generating-a-summary-document-i-am-given-suggestions-of-valid-email-address-formats-when-entering-invalid-e)

PR #289, opened 2017-06-19
Add organisation id to csv export

jongilbraith:
Surfacing the `organisation_content_id` of the user who generated the summary under the column `content_id`.

### DIFF
--- a/app/services/csv_generator.rb
+++ b/app/services/csv_generator.rb
@@ -22,10 +22,7 @@ class CsvGenerator
   private
 
   def row(record)
-    record
-      .attributes
-      .slice(*attributes)
-      .map { |key, value| format(key, value) }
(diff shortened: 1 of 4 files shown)


app/services/csv_generator.rb
```diff
@@ -22,10 +22,7 @@ def attributes
   private
 
   def row(record)
-    record
-      .attributes
-      .slice(*attributes)
-      .map { |key, value| format(key, value) }
+    attributes.map { |attribute| format(attribute, record.send(attribute)) }
```
benlovell: You may well be able to `record[attribute]` here instead of `send`. If not, can we use `public_send` instead?

jongilbraith:
[As an analyst I want to know which delivery partner generated a summary document](https://trello.com/c/XnS6qU9K/202-as-an-analyst-i-want-to-know-which-delivery-partner-generated-a-summary-document)

PR #290, opened 2017-06-21
Fix click to fill behaviour on email suggestion

jongilbraith:
Was positive I recalled testing this but apparently not.

Picked up by Aman yesterday, the ability to click the suggested email to


benlovell:
Good catch! 🏆

PR #301, opened 2018-03-07
Bump templates for MAS number change

benlovell:


benlovell:
[Ensure summary documents contain the correct telephone number for MAS](https://trello.com/c/vKSkj155/430-ensure-summary-documents-contain-the-correct-telephone-number-for-mas)

PR #307, opened 2018-06-05
Switch to our fork temporarily

benlovell:
We should move away from the multipack and configure our buildpacks
individually. I'll do this shortly.

### DIFF
--- a/app.json
+++ b/app.json
@@ -37,7 +37,7 @@
       "url": "heroku/ruby"
     },
     {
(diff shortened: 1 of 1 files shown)


benlovell:
[Update summary documents to correct typo on 'Take Cash in Chunks' section](https://trello.com/c/qko2uHum/558-update-summary-documents-to-correct-typo-on-take-cash-in-chunks-section)

PR #308, opened 2018-06-19
Bump Ruby and gem dependencies

benlovell:


benlovell:
[Allow both phone and face-to-face appointments to be summarised by the same user account](https://trello.com/c/xFTC4HPH/584-allow-both-phone-and-face-to-face-appointments-to-be-summarised-by-the-same-user-account)

PR #311, opened 2018-10-01
Correct issue with disabled/readonly field

benlovell:
When the field is marked with the `disabled` attribute it does not POST
a value, thus would be preselected as `standard` for the
`appointment_type`. This means some customers were missing out on vital
information pertaining to the pensionable age.

I also scoped the test helper methods as they were clobbering eachother
in the global namespace.


benlovell:
[Investigate possible issue with summary document](https://trello.com/c/mUcLaCnL/703-investigate-possible-issue-with-summary-document)

PR #312, opened 2018-10-02
Validate email presence when requesting digital

benlovell:
Digital summaries should not be created when no email address was
specified.


benlovell:
[As a guider, I should not be able to generate a digital summary document without an email address](https://trello.com/c/nM7GFB9y/693-as-a-guider-i-should-not-be-able-to-generate-a-digital-summary-document-without-an-email-address)

PR #316, opened 2019-02-14
Work around missing users

benlovell:
There are some instances of old data where the associated user is
missing or not synchronised from signon. This ensures the organisation
ID is accessed safely in those instances.


benlovell:
[Ensure CSVs can be exported in full](https://trello.com/c/fN4E0GRI/895-ensure-csvs-can-be-exported-in-full)

PR #327, opened 2019-12-11
Update output templates for signposting

benlovell:
A calculator was inaccurate so was replaced by MAS. This update ensures
we point to the correct version.


benlovell:
[Signpost to the correct MAS pension draw down calculator](https://trello.com/c/Kw9TpgQn/1225-signpost-to-the-correct-mas-pension-draw-down-calculator)

PR #342, opened 2020-05-06
Adds further reporting filters

benlovell:
Adds the telephone and requested digital filters to the admin reporting.
They both default to unspecified so will include all results unless a
particular value is selected.


benlovell:
[Add channel filter to summary document reporting](https://trello.com/c/tdEVgVZ2/1425-add-channel-filter-to-summary-document-reporting)

